Issue #20, opened 2023-06-16
Instr/IndexOf support BinaryString as input parameter


pengzhiwei2018:
Currently GeaFlow uses BinaryString as its implementation for string, which provides better performance. However, at present, the built-in String Function uses string as the input parameter, and a direct conversion from BinaryString to String is required in the calculation phase, which has performance overhead. Implement the InStr/IndexOf function with BinaryString as an input.

liuxiaocs7:
Hi, @pengzhiwei2018 and @Leomrlin, this issue can be closed by #32

PR #32, opened 2023-06-20
[#20]Instr/IndexOf support BinaryString as input parameter

Kwafoor:
Implemented the indexOf function in BinaryString by referencing spark UTF8String. Instr/IndexOf support BinaryString as input parameter.

### How was this PR tested?

### DIFF
--- a/geaflow/geaflow-common/src/main/java/com/antgroup/geaflow/common/binary/BinaryString.java
+++ b/geaflow/geaflow-common/src/main/java/com/antgroup/geaflow/common/binary/BinaryString.java
@@ -226,6 +226,33 @@ public class BinaryString implements Comparable<BinaryString>, Serializable, Kry
         return false;
     }
 
+    public int indexOf(BinaryString v, int start) {
+        if (v.numBytes == 0) {
+            return 0;
+        }
+
+        // locate to the start position.
+        int i = 0; // position in byte
+        int c = 0; // position in character
+        while (i < numBytes && c < start) {
+            i += numBytesForFirstByte(getByte(i));
+            c += 1;
+        }
+
+        do {
+            if (i + v.numBytes > numBytes) {
+                return -1;
+            }
+            if (BinaryOperations.arrayEquals(binaryObject, offset + i, v.binaryObject, v.offset, v.numBytes)) {
+                return c;
+            }
+            i += numBytesForFirstByte(getByte(i));
+            c += 1;
+        } while (i < numBytes);
+
+        return -1;
+    }
+
     public boolean matchAt(final BinaryString s, int pos) {
         if (s.numBytes + pos > numBytes || pos < 0) {
             return false;
--- a/geaflow/geaflow-common/src/main/java/com/antgroup/geaflow/common/binary/BinaryString.java
+++ b/geaflow/geaflow-common/src/main/java/com/antgroup/geaflow/common/binary/BinaryString.java
@@ -226,8 +226,8 @@ public class BinaryString implements Comparable<BinaryString>, Serializable, Kry
         return false;
     }
 
-    public int indexOf(BinaryString v, int start) {
-        if (v.numBytes == 0) {
+    public int indexOf(BinaryString s, int start) {
+        if (s.numBytes == 0) {
             return 0;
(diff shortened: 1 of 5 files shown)


pengzhiwei2018:
@Leomrlin  can you take a look at this PR first?

geaflow/geaflow-dsl/geaflow-dsl-plan/src/main/java/com/antgroup/geaflow/dsl/udf/table/string/Instr.java
```diff
@@ -44,4 +45,29 @@ public Long eval(String str, String target, Long from, Long nth) {
...
+
+    public Long eval(BinaryString str, BinaryString target, Long from, Long nth) {
+        if (str == null || target == null || from == null || nth == null) {
+            return null;
+        }
+        if (nth <= 0) {
+            return null;
+        }
+        int fromIndex = from.intValue() - 1;
+        if (fromIndex < 0) {
```
Leomrlin, line 65: Why do we need to return null here?

Leomrlin (approved):
It looks good! Thank you for adding relevant comments and a lot of test cases. However, there is a small issue that may require your feedback.

geaflow/geaflow-common/src/main/java/com/antgroup/geaflow/common/binary/BinaryString.java
```diff
@@ -226,6 +226,33 @@ public boolean contains(final BinaryString substring) {
         return false;
     }
 
+    public int indexOf(BinaryString v, int start) {
```
pengzhiwei2018: Maybe the variable "s" or "str" is more reasonable than "v"

pengzhiwei2018 (commented):
@Kwafoor Thanks for you contribution. Nice work! After some minor comments fixed, we can merge it.

geaflow/geaflow-dsl/geaflow-dsl-plan/src/main/java/com/antgroup/geaflow/dsl/udf/table/string/Instr.java
beryllw: > Why do we need to return null here?

How about returning -1 to be consistent with `IndexOf`? Returning null is consistent with the previous code `Instr`.

geaflow/geaflow-common/src/main/java/com/antgroup/geaflow/common/binary/BinaryString.java
beryllw: Ok, I'll make the change.

geaflow/geaflow-dsl/geaflow-dsl-plan/src/main/java/com/antgroup/geaflow/dsl/udf/table/string/Instr.java
Leomrlin: There may be multiple interpretations here. My understanding is that fromIndex < 0 can be regarded as starting the search from the beginning of the string. However, in Aliyun's ODPS, a negative start_position is considered to indicate that the starting position is counted backwards from the end of the string. Refer to https://help.aliyun.com/document_detail/455546.html?spm=a2c4g.48973.0.i1 for more information.

pengzhiwei2018: @Kwafoor @Leomrlin The function **Instr** is defined the Hive SQL and most SQL Engine on BigData adopt it. So we should also follow the semantics of hive(ODPS also follow this semantics).   I will do  this in another PR. So we can stop the arguing for `fromIndex < 0` here.

Leomrlin: ok, that's great. It's not a problem.

pengzhiwei2018 (approved):
LGTM

Issue #5, opened 2023-06-13
Support create temporary table/graph


pengzhiwei2018:
Support create temporary table or graph which will store in the memory. Currently geaflow only support create persistent table or graph which will store the meta to console catalog. However it is not flexible to modify the config on the table. So we need support temporary table or graph.

Kwafoor:
Hi, @pengzhiwei2018 , this issue can be closed by #46

PR #46, opened 2023-06-25
Support create temporary table/graph #5

Leomrlin:
Support create temporary table or graph which will store in the memory. Currently geaflow only support create persistent table or graph which will store the meta to console catalog. However it is not flexible to modify the config on the table. So we need support temporary table or graph.

In addition, support using statements in the DDL graph definition, which are automatically transformed into insert statements when executed. If all nodes and edges are ALL_WINDOW, the graph will be optimized and stored as a static graph.

### How was this PR tested?


pengzhiwei2018 (approved):
LGTM

PR #66, opened 2023-06-30
Fix status to be finished when batch task completed

fanzhidongyzby:
Refrence: https://github.com/TuGraph-family/tugraph-analytics/issues/65


pengzhiwei2018 (approved):
LGTM

PR #86, opened 2023-07-11
[ISSUE-82] Support type registration optimization for Kryo serializer

Leomrlin:
GeaFlow uses Kryo as the default serializer, but some data types in the current DSL module, such as Path, Message, and Row, have not been registered in Kryo, which may lead to a decrease in the performance of the serializer. This pr supports registration optimization for these types.

### How was this PR tested?


pengzhiwei2018 (approved):
LGTM

PR #128, opened 2023-08-18
[ISSUE-127] Support stream distinct

Leomrlin:
The existing distinct operator is transformed into an aggregation operator, which only aggregates data from a single window and outputs the result at the finish of each window. This results in duplicates in distinct calculations for streams with multiple windows, as the same data item is output in multiple windows. There is a need to support distinct across windows in the stream, ensuring that the same data item is only output once during the complete computation of the data stream.

This PR adds a dedicated Aggregate functions for Distinct, modifying the aggregation results to only have a value when first read, and subsequently returning null values to eliminate duplicate results.

### How was this PR tested?


pengzhiwei2018 (approved):
LGTM

PR #133, opened 2023-08-22
[ISSUE-132] Do checkpoint after single window job finished

cbqiao:
A job will do checkpoint every 5 window by default. However a single window job should do checkpoint and finish after the only window done.

### How was this PR tested?


pengzhiwei2018 (approved):
LGTM

PR #135, opened 2023-08-22
[ISSUE-134] Simplify the static graph semantics in DSL

Leomrlin:
The current DSL supports two types of semantics, static graph and dynamic graph, which can easily confuse users. For simplicity, it is necessary to simplify and unify the static graph semantics.
Firstly, for a given vertex-edge stream that forms a graph using PStreams, it is considered a static graph, which is only stored in memory.
Secondly, when querying a snapshot of the dynamic graph at a specific moment, although the graph itself is dynamic, the computation is performed statically on the snapshot.
Except for these two cases, all other graphs are considered dynamic graphs.

### How was this PR tested?


pengzhiwei2018 (approved):
LGTM

PR #140, opened 2023-08-28
[ISSUE-139] Fix  NPE for UDGA when the target vertex  is not exist

Leomrlin:
In dynamic graph algorithms, the process() method is called without considering whether the vertices exist. If a vertex has not yet been inserted into the graph and is accessed, it will cause an NPE (NullPointerException), such as when accessing the endpoint of an edge. It is necessary to add null check logic to skip the target vertex that is null.

### How was this PR tested?


pengzhiwei2018 (approved):
LGTM

PR #146, opened 2023-08-30
[ISSUE-145] Fix exception for geaflow plugin type when using JDBC connector

Leomrlin:
Create table in console catalog report plugin type not found when using JDBC as the connector type. The reason is that the catalog creates an enumeration object of the plugin type using the connector name, but the JDBC type does not exist in the code. In order to simplify the process of adding different connector types in the future and avoid similar issues, the enumeration class has been removed and replaced with passing a string directly.

### How was this PR tested?


pengzhiwei2018 (approved):
LGTM

PR #150, opened 2023-09-01
[ISSUE-149] Fix NPE for pluginType check

652053395:
1. OSS URL format
2. pluginType null check
3. validate edge/vertex fields when create


pengzhiwei2018 (approved):
LGTM

PR #154, opened 2023-09-01
[ISSUE-153] Support translate sql join to graph match

Leomrlin:
For users who are not familiar with the GQL language, it can be difficult to directly use Match to process graphs. GeaFlow aims to materialize joins using graphs and utilize Match to achieve join calculations indirectly. Therefore, we hope that users can write Join queries in a SQL-like manner, and GeaFlow DSL will automatically convert them into GQL processing. This syntactical transformation not only guarantees performance but also provides convenience to users, allowing them to enjoy the benefits of graph computing without writing any GQL syntax.

A Join query that can be transformed into graph computation is similar to the following example:

INSERT INTO tbl_result
SELECT s.id, sc.targetId, c.name, hasTeacher.targetId
FROM student s, selectCourse sc, course c, hasTeacher
WHERE s.id = sc.srcId AND c.id = sc.targetId AND c.id = hasTeacher.srcId
;

Of course, the prerequisite for this transformation is to have a clear graph schema. Users still need to define the graph schema in advance and declare the graph used for the query.

CREATE GRAPH IF NOT EXISTS g_student (
Vertex student (
id bigint ID,
name varchar,
age int
),
Vertex course (
id bigint ID,
name varchar,
course_hour int
),
Vertex teacher (
id bigint ID,
name varchar
),
Edge selectCourse (
srcId from student SOURCE ID,
targetId from course DESTINATION ID,
ts bigint
),
Edge hasTeacher (
srcId from course SOURCE ID,
targetId from teacher DESTINATION ID
)
) WITH (
storeType='rocksdb',
shardCount = 2
);

USE GRAPH g_student;

### How was this PR tested?


pengzhiwei2018 (approved):
LGTM

PR #159, opened 2023-09-05
[ISSUE-158] JDBC connector cannot write data into mysql database

Leomrlin:
 After using the JDBC Connector to connect to MYSQL for writing, the job runs successfully but the data is not written to the database. The reason is that the Sink implementation creates a temporary table, causing the data to be written into the temporary table.

### How was this PR tested?


pengzhiwei2018 (approved):
LGTM

PR #163, opened 2023-09-11
[ISSUE-164] Add keyby for sink table with primary key

pengzhiwei2018:

For sink table with primary key, e.g. mysql, we need do data shuffle by the primary key before write data to the sink table.
If not do this,  write conflict will happen on the same primary key when  different task update data to the same key.
### How was this PR tested?


Leomrlin (approved):
LGTM

PR #165, opened 2023-09-14
[ISSUE-166] Optimize logical plan for test case

pengzhiwei2018:

Currently we add a graph define for each test case by default. It will make the logical plan complex for the case which do not need graph.
This PR remove the default graph define and add graph define for the test case which really need.
### How was this PR tested?


Leomrlin (approved):
LGTM

PR #180, opened 2023-09-26
[ISSUE-179] Support converting traversal all to traversal by stream

Leomrlin:
Support converting traversal all vertices in the graph to traversal by stream execution. Traversal by stream can divide trigger points into several groups to reduce the number of trigger points in each group and lower the peak memory usage during execution. This can effectively reduce the possibility of GC or OOM, but the trade-off is that the runtime cost may increase rapidly.

### How was this PR tested?


pengzhiwei2018 (approved):
LGTM

PR #200, opened 2023-10-17
[ISSUE-199] Fix known issues with DSL and optimize LDBC test

Leomrlin:
Fix the index calculation error in the loopUtil operator.
Fix the termination iteration calculation for loopUtil.
Fix the issue where configuration options created through Console are not mapped with short names.
Fix the problem where running SQL without Create Graph does not trigger graph meta information callback.

### How was this PR tested?


pengzhiwei2018 (approved):
LGTM

PR #215, opened 2023-10-26
[ISSUE-214] Support different names for vertex and edge meta fields 

Leomrlin:
Currently, due to the presence of the union operation, the constraint for the meta fields of graph vertices and edges is that all vertex IDs must have the same name, and all edge source, destination, and timestamps must have the same name. This constraint is clearly too strong, and it is necessary to support meta fields with different names.

### How was this PR tested?


pengzhiwei2018 (approved):
LGTM

PR #228, opened 2023-10-31
[ISSUE-227] Drivers and containers check and register heartbeat after failover

dy247846795:
* Drivers and containers now check and register heartbeat after failover of master.
* Add labels for k8s services and deployments.

### How was this PR tested?


pengzhiwei2018 (approved):
LGTM

PR #235, opened 2023-11-01
[ISSUE-234] Support convert Left Join to Graph Match 

Leomrlin:
In Issue197(https://github.com/TuGraph-family/tugraph-analytics/issues/197), we support translating the SQL join and other operators into GQL graph matching, but currently only support inner join. However, for graph traversal, left join can also be effectively transformed into graph match execution. Therefore, we can extend the support for converting Left Join into graph match.

### How was this PR tested?


pengzhiwei2018 (approved):
LGTM

PR #250, opened 2023-11-07
[ISSUE-249] Fix redis plugin in geaflow-console

dy247846795:
1. Support more Redis params such as  username, password, timeout and retry.
2. Now using jedis-pool to replace jedis, to avoid unsafety in multi-thread case.


Leomrlin (approved):
LGTM

PR #258, opened 2023-11-14
[ISSUE-217] Support dynamic graph incremental algorithm

Leomrlin:
The current dynamic graph algorithm in DSL actually runs on all vertices in the graph. Based on user feedback, it is necessary to support the incremental graph algorithm for dynamic graphs.

We have designed this feature as follows:

The set of inserted vertices, source vertices of inserted edges, and destination vertices of inserted edges belonging to each window of the dynamic graph are used as trigger vertices for the first round of algorithm iteration.
Distinguish between the concept of vertex data in the graph and the data updated by the algorithm on vertices, so that users can use both during algorithm implementation.
When the algorithm finishes, the set of trigger vertices and all vertices that have received messages in algorithm iterations are used as the callback objects for the finish() interface.
After each dynamic graph window ends, the updated data on the vertices in the graph is persisted to disk, making it convenient to continue running the graph algorithm from the interrupted position when the job restarts.

We have added a configuration option to control the activation of the incremental graph algorithm functionality. Admittedly, implementing incremental graph algorithms imposes higher requirements on UDGA. Not all algorithms can be implemented as incremental graph algorithms, and the implementation of incremental UDGA may differ slightly from that of regular algorithm UDGA. Users can enable this feature according to their usage scenarios.

### How was this PR tested?


pengzhiwei2018 (approved):
LGTM

PR #267, opened 2023-11-20
[ISSUE-264] add supervisor to manager worker process lifecycle

qingwen220:
This PR aims to improve the failover efficiency in k8s.

Currently, the cluster will rebuild the whole cluster by deleting all pods and recreating them when a runtime exception is thrown. This process will last several minutes or more when the cluster is resource-limited.

By incurring the supervisor, we could restart the worker process without releasing the pod resource.


pengzhiwei2018 (approved):
LGTM

PR #268, opened 2023-11-21
[ISSUE-263]: add a new graph store

qingwen220:
CStore is a cloud-friendly graph store that stores graph data either locally or remotely (such as OSS, and HDFS).


pengzhiwei2018 (approved):
LGTM

PR #315, opened 2023-12-26
[ISSUE-314] add event listener for scheduler.

cbqiao:
register event listener when submit event to worker
then received response will dispatch to event listener to handle.


pengzhiwei2018 (approved):
LGTM

PR #322, opened 2024-01-16
[ISSUE-321]Added redis service registration and discovery, Optimizing the jdbc Protocol

KevinLi724:
1.Service registration and discovery based on redis is more lightweight.
2.Tugraph analysis server optimization query metadata compilation.
3.Optimize the jdbc protoco.


pengzhiwei2018 (approved):
LGTM

PR #342, opened 2024-03-11
[ISSUE-341] Support model inference

KevinLi724:
This submission is mainly about the ability of graph computing to integrate AI model inference, the first is to support the custom virtual environment at the job level; Based on the ability of shared memory to communicate across processes, model inference interaction between java processes and python inference processes is supported.


pengzhiwei2018 (approved):
LGTM

PR #345, opened 2024-03-19
Support query and analytics on graph by natural language

652053395:
Use LLM to realize Text2GQL，and provide LLM register and natureal language query on console.


fanzhidongyzby (approved):
LGTM

PR #378, opened 2024-11-19
[ISSUE-377] Fix demo job and auto create demo job when console starts

652053395:
Fix demo job and auto create demo job when console starts


Leomrlin (approved):
LGTM

PR #380, opened 2024-11-19
[ISSUE-379] Fix the issue of the demo case getting stuck

QiZhang1997:
Optimized the SQL in the demo case by adding data validation logic to ensure that erroneous data will not be transmitted to the engine.


Leomrlin (approved):
LGTM

PR #491, opened 2025-04-01
The submission of two gql files simultaneously using `gql_submit` failed due to a port conflict related to the dashboard. 

Leomrlin:
The submission of two gql files simultaneously using `gql_submit` failed due to a port conflict related to the dashboard.


Loognqiang (approved):
LGTM

PR #509, opened 2025-04-23
[STORE]Add paimon graph table handle.

tanghaodong25:
This is part of the code that supports the Paimon job graph data format State, used to replace the previous KV-based Paimon Table.


Loognqiang (approved):
LGTM

PR #550, opened 2025-07-10
Add random table source

Leomrlin:
Add random table source

example:

CREATE TABLE tbl_person (
	id bigint,
	name varchar,
	age int
) WITH (
	type='random',
	`geaflow.dsl.window.size`=20,
	`geaflow.dsl.random.source.tuples.per.second`=100.0,
	`geaflow.dsl.random.source.max.batch`=100
);


Loognqiang (approved):
LGTM

PR #578, opened 2025-08-04
Support ASync-Write mode for PaimonGraphStore

cfsfine:
Support ASync-Write mode for PaimonGraphStore.

### How was this PR tested?
- [✅] Tests have Added for the changes
- [❎] Production environment verified


geaflow/geaflow-plugins/geaflow-store/geaflow-store-paimon/src/main/java/org/apache/geaflow/store/paimon/proxy/AsyncPaimonGraphRWProxy.java
```diff
@@ -0,0 +1,113 @@
...
+
+    public AsyncPaimonGraphRWProxy(PaimonTableRWHandle vertexHandle, PaimonTableRWHandle edgeHandle,
+                                   int[] projection, IGraphKVEncoder<K, VV, EV> encoder,
+                                   Configuration config) {
+        super(vertexHandle, edgeHandle, projection, encoder);
+        this.flushBuffer = new AsyncFlushBuffer<>(config, this::flush,
+            SerializerFactory.getKryoSerializer());
+    }
+
+    private void flush(GraphWriteBuffer<K, VV, EV> graphWriteBuffer) {
```
Loognqiang, line 51: When will this private interface be used? And why is this private interface needed when has public flush() interface.

cfsfine: > When will this private interface be used? And why is this private interface needed when has public flush() interface.

As you can see, the public method flush is exposed to out. While the private method flush is passed into the internal member `AsyncBuffer` as a callback function, which will be called when the `AsyncBuffer` is persisted to disk.

Loognqiang: I got it.

tanghaodong25:
LGTM

PR #585, opened 2025-08-11
[asf] Remove publishing profile from .asf.yaml

vesense:
Remove publishing profile from .asf.yaml since we already have a separate website repo.


Loognqiang (approved):
LGTM

PR #608, opened 2025-09-03
refactor : pipeline async strategy

kitalkuyo-gita:
### How was this PR tested?

### DIFF
--- a/geaflow/geaflow-core/geaflow-engine/geaflow-cluster/src/main/java/org/apache/geaflow/cluster/client/strategy/ResourceBasedAsyncStrategy.java
+++ b/geaflow/geaflow-core/geaflow-engine/geaflow-cluster/src/main/java/org/apache/geaflow/cluster/client/strategy/ResourceBasedAsyncStrategy.java
@@ -1,0 +1,77 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.geaflow.cluster.client.strategy;
+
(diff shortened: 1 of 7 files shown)


geaflow/geaflow-core/geaflow-engine/geaflow-cluster/src/main/java/org/apache/geaflow/cluster/client/strategy/ResourceBasedAsyncStrategy.java
```diff
@@ -0,0 +1,77 @@
...
+        int containerNum = config.getInteger(ExecutionConfigKeys.CONTAINER_NUM);
+        
+        // Get configurable thresholds, fall back to defaults if not specified
+        int driverThreshold = config.getInteger(
+            FrameworkConfigKeys.ASYNC_RESOURCE_DRIVER_THRESHOLD, DEFAULT_ASYNC_THRESHOLD_DRIVERS);
+        int containerThreshold = config.getInteger(
+            FrameworkConfigKeys.ASYNC_RESOURCE_CONTAINER_THRESHOLD, DEFAULT_ASYNC_THRESHOLD_CONTAINERS);
+        
+        // Use async mode if we have multiple drivers or many containers
+        boolean shouldUseAsync = driverNum >= driverThreshold || containerNum >= containerThreshold;
```
cbqiao, line 51: Why is asynchronous mode enabled when there are multiple drivers or containers, especially when the number of containers exceeds the threshold? It seems that the original logic does not use this condition to trigger async mode.

kitalkuyo-gita: Hi.[cbqiao](https://github.com/cbqiao). I'm glad you reviewed my code. The variable FrameworkConfigKeys#ASYNC_MODE_STRATEGY provides three modes, among which config_based is the original mode. In addition, the other two modes are provided. They are developed because of the characteristics of asynchronous mode. The original intention of the design of asynchronous mode is to make more full use of cluster resources. When the number of drivers is large or the number of containers reaches a certain scale, the concurrency and complexity of the system will increase. At this time, the synchronous mode often becomes a bottleneck, resulting in insufficient utilization of resources. The asynchronous mode can better decouple the dependencies between various computing units and improve overall throughput and resource utilization. In actual production environments, the number of drivers and containers are important indicators for measuring the scale of jobs. As long as either party reaches the asynchronous threshold, it means that the job already has a high concurrency, and the use of asynchronous mode can significantly improve operating efficiency. If only the number of drivers is used as a trigger condition, many large-scale container scenarios will be ignored, resulting in performance waste. So I think it is necessary

cbqiao: @Loognqiang Async mode is commonly used in multi-driver scenarios. Please help check if the change makes sense.

PR #618, opened 2025-09-12
Jdk11 ci support

Loognqiang:
### How was this PR tested?
#601

### DIFF
--- a/geaflow/geaflow-dsl/geaflow-dsl-runtime/pom.xml
+++ b/geaflow/geaflow-dsl/geaflow-dsl-runtime/pom.xml
@@ -154,7 +154,7 @@
             <plugin>  
                 <groupId>org.apache.maven.plugins</groupId>  
                 <artifactId>maven-surefire-plugin</artifactId>  
-                <version>3.0.0-M7</version>
+                <version>${maven-surefire.version}</version>
                 <configuration>  
                     <!-- Optimize memory usage -->  
                     <argLine>-Xmx512m -XX:MaxMetaspaceSize=256m -XX:+UseG1GC -XX:MaxGCPauseMillis=200 -Djava.awt.headless=true</argLine>  
@@ -190,7 +190,7 @@
                     <plugin>  
                         <groupId>org.apache.maven.plugins</groupId>  
                         <artifactId>maven-surefire-plugin</artifactId>  
-                        <version>3.0.0-M7</version>
+                        <version>${maven-surefire.version}</version>
                         <configuration>  
                             <argLine>-Xmx1024m -XX:MaxMetaspaceSize=512m -XX:+UseG1GC</argLine>  
                             <forkCount>1</forkCount>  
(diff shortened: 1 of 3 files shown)


.github/workflows/ci-jdk11.yml
```diff
@@ -0,0 +1,68 @@
...
+      - name: Build and Test On JDK 11
+        run: |
+          test_modules="!geaflow/geaflow-dsl/geaflow-dsl-connector/geaflow-dsl-connector-hive,"
+          test_modules+="!geaflow/geaflow-dsl/geaflow-dsl-runtime,"
+          test_modules+="!geaflow/geaflow-analytics-service/geaflow-analytics-service-common,"
+          test_modules+="!geaflow/geaflow-analytics-service/geaflow-analytics-service-client,"
+          test_modules+="!geaflow/geaflow-analytics-service/geaflow-analytics-service-server,"
+          test_modules+="!geaflow/geaflow-examples,"
+          test_modules+="!geaflow/geaflow-deploy/geaflow-assembly,"
+          test_modules+="!geaflow-mcp"
```
cbqiao, line 67: Are all these modules unable to run on JDK 11? Especially the geaflow-examples module, does this mean none of the example cases support running on JDK 11 yet?

geaflow/geaflow-dsl/geaflow-dsl-runtime/pom.xml
```diff
@@ -154,7 +154,7 @@
             <plugin>  
                 <groupId>org.apache.maven.plugins</groupId>  
                 <artifactId>maven-surefire-plugin</artifactId>  
-                <version>3.0.0-M3</version>
```
qingwen220, line 157: this version is supposed to be define in the parent pom.xml

yaozhongq (approved):
LGTM

geaflow/geaflow-dsl/geaflow-dsl-runtime/pom.xml
Loognqiang: I have addressed it.

geaflow/geaflow-common/src/main/java/org/apache/geaflow/common/binary/BinaryOperations.java
```diff
@@ -21,11 +21,10 @@
 
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
-import sun.misc.Unsafe;
 
 public class BinaryOperations {
 
-    private static final Unsafe UNSAFE;
+    private static final sun.misc.Unsafe UNSAFE;
```
cbqiao, line 27: The sun.misc.Unsafe class was removed in JDK 9. Can the code still run properly on JDK 11?

Loognqiang: The sun.misc.Unsafe class is present in JDK 11 but belongs to the jdk.unsupported module. We can allow the current code to run on JDK 11 by adding specific JVM arguments. For JDK 17+ versions, we will directly adopt new standard APIs.

.github/workflows/ci-jdk11.yml
Loognqiang: Following issue #619 , we will refactor the Connector unit tests and dependencies. The examples module will no longer be affected.

cbqiao (approved):
LGTM

PR #621, opened 2025-09-16
[ISSUE-620] fix: remove the directory of db after running PaimonRWHandle

dy247846795:
[ISSUE-620] fix: remove the directory of db after running PaimonRWHandle

### How was this PR tested?
- [yes] Tests have Added for the changes
- [no] Production environment verified

### DIFF
--- a/geaflow/geaflow-plugins/geaflow-store/geaflow-store-paimon/src/test/java/org/apache/geaflow/store/paimon/PaimonRWHandleTest.java
+++ b/geaflow/geaflow-plugins/geaflow-store/geaflow-store-paimon/src/test/java/org/apache/geaflow/store/paimon/PaimonRWHandleTest.java
@@ -65,7 +65,7 @@ public class PaimonRWHandleTest {
 
         @Override
         public void drop() {
-
+            super.drop();
         }
     }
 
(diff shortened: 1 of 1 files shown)


geaflow/geaflow-plugins/geaflow-store/geaflow-store-paimon/src/test/java/org/apache/geaflow/store/paimon/PaimonRWHandleTest.java
```diff
@@ -77,7 +77,8 @@ public void testGraphStore() throws Exception {
         config.put(ExecutionConfigKeys.JOB_APP_NAME, "test_paimon_app");
         storeBase.init(new StoreContext("test_paimon_store").withConfig(config));
 
-        PaimonTableRWHandle edgeHandle = storeBase.createEdgeTableHandle(new Identifier("test_paimon_store", "edge"));
+        PaimonTableRWHandle edgeHandle = storeBase.createEdgeTableHandle(
```
Loognqiang, line 80: The context in this line passes in the name "test_paimon_store":
storeBase.init(new StoreContext("test_paimon_store").withConfig(config));
On line 81, the first parameter of the Identifier constructor is changed to storeBase.paimonStoreName. These two should be consistent.

dy247846795: The actual db_name will be changed to "test_paimon_store#0" during Init, which is already consistent in StoreBase. So "test_paimon_store" is only used once and not needed anymore.

PR #623, opened 2025-09-16
[ISSUE-622] feat: support distributed writing with one Paimon graph store

### LINKED ISSUE #622: Support loading data into graph store with Paimon type in distributed mode
**Is your feature request related to a problem? Please describe.**
DSL supports loading data from source tables into a distributed Paimon graph store within a single database.

qingwen220:
Supports a new Graphview type Paimon. e.g., 
```
CREATE GRAPH dy_modern (
	Vertex person (
	  id bigint ID,
	  name varchar,
	  age int
	),
	Edge knows (
	  srcId bigint SOURCE ID,
	  targetId bigint DESTINATION ID,
	  weight double
	)
) WITH (
	storeType='paimon',
	shardCount = 1
);
```

resolve #622 ISSUE-622


geaflow/geaflow-core/geaflow-runtime/geaflow-operator/src/main/java/org/apache/geaflow/operator/impl/graph/materialize/GraphViewMaterializeOp.java
```diff
@@ -105,12 +113,28 @@ public void checkpoint(long windowId) {
...
         this.graphState.manage().operate().archive();
+
+        if (graphViewDesc.getBackend() == BackendType.Paimon) {
+            int taskIndex = runtimeContext.getTaskArgs().getTaskIndex();
+
+            PaimonCommitRegistry registry = PaimonCommitRegistry.getInstance();
+            List<TaskCommitMessage> messages = registry.pollMessages(taskIndex);
+            if (messages != null && !messages.isEmpty()) {
+                for (TaskCommitMessage message : messages) {
+                    List<CommitMessage> msg = message.getMessages();
```
kitalkuyo-gita, line 124: message.getMessages() may also return null, so it's recommended to add a null check.

geaflow/geaflow-core/geaflow-runtime/geaflow-operator/src/main/java/org/apache/geaflow/operator/impl/graph/materialize/GraphViewMaterializeOp.java
```diff
@@ -105,12 +113,28 @@ public void checkpoint(long windowId) {
...
+                    LOGGER.info("task {} emits windowId:{} chkId:{} table:{} messages:{}",
+                        taskIndex, windowId, checkpointId, message.getTableName(),
+                        msg.size());
+                    collectValue(new PaimonMessage(checkpointId, message.getTableName(),
```
kitalkuyo-gita, line 128: Is it allowed to pass an empty list to collectValue? Will it cause downstream errors?
eq
```
List<CommitMessage> msg = message.getMessages();
            ...
            collectValue(new PaimonMessage(checkpointId, message.getTableName(),
                msg));
```

geaflow/geaflow-core/geaflow-runtime/geaflow-operator/src/main/java/org/apache/geaflow/operator/impl/graph/materialize/PaimonGlobalSink.java
```diff
@@ -0,0 +1,138 @@
...
+            return;
+        }
+
+        final long startTime = System.currentTimeMillis();
+        final long checkpointId = paimonMessages.get(0).getCheckpointId();
+        Map<String, List<CommitMessage>> tableMessages = new HashMap<>();
+        for (PaimonMessage message : paimonMessages) {
+            List<CommitMessage> messages = message.getMessages();
+            if (messages != null && !messages.isEmpty()) {
+                String tableName = message.getTableName();
```
kitalkuyo-gita, line 87: For multiple messages with the same tableName, there may be merging, duplication, and order requirements, and it is necessary to ensure that they are not lost or confused.

geaflow/geaflow-core/geaflow-runtime/geaflow-operator/src/main/java/org/apache/geaflow/operator/impl/graph/materialize/PaimonGlobalSink.java
```diff
@@ -0,0 +1,138 @@
...
+        }
+        LOGGER.info("committed chkId:{} messages:{} deserializeCost:{}ms",
+            checkpointId, paimonMessages.size(), deserializeTime);
+        paimonMessages.clear();
+    }
+
+    private StreamWriteBuilder getWriteBuilder(String tableName) {
+        return writeBuilders.computeIfAbsent(tableName, k -> {
+            try {
+                FileStoreTable table = (FileStoreTable) client.getTable(Identifier.create(dbName,
```
kitalkuyo-gita, line 121: If client.getTable throws an exception, it will be wrapped and thrown; however, if the table does not exist and auto-create is not enabled, an error will be reported directly. It is recommended to provide clear exception information.

geaflow/geaflow-plugins/geaflow-store/geaflow-store-paimon/src/main/java/org/apache/geaflow/store/paimon/BasePaimonGraphStore.java
```diff
@@ -19,27 +19,75 @@
...
 
     @Override
     public IFilterConverter getFilterConverter() {
         return filterConverter;
     }
+
+    protected PaimonTableRWHandle createVertexTable(int shardId) {
+        String tableName = vertexTable;
+        if (StringUtils.isEmpty(tableName)) {
+            tableName = String.format("%s#%s", "vertex", shardId);
```
kitalkuyo-gita, line 68: - When reading the config, if no configuration is found, a fallback is executed. However, is the fallback string consistent with other locations? (For example, is the tableName concatenation format consistent?)

geaflow/geaflow-plugins/geaflow-store/geaflow-store-paimon/src/main/java/org/apache/geaflow/store/paimon/PaimonTableRWHandle.java
```diff
@@ -63,13 +80,33 @@ public void write(GenericRow row, int bucket) {
...
-        } catch (Exception e) {
-            throw new GeaflowRuntimeException("Failed to commit data into Paimon.", e);
+        commit(checkpointId, false);
+    }
+
+    public void commit(long checkpointId, boolean waitCompaction) {
+        flush(checkpointId, waitCompaction);
+        List<CommitMessage> messages = new ArrayList<>();
+        for (CommitMessage commitMessage : commitMessages) {
+            if (((CommitMessageImpl) commitMessage).isEmpty()) {
```
kitalkuyo-gita: Force the type conversion to CommitMessageImpl. If it's not of this type, an exception will be thrown. Ensure that all CommitMessages implement this implementation, or add an instanceof check.

geaflow/geaflow-plugins/geaflow-store/geaflow-store-paimon/src/main/java/org/apache/geaflow/store/paimon/PaimonTableRWHandle.java
```diff
@@ -63,13 +80,33 @@ public void write(GenericRow row, int bucket) {
...
-            flush(checkpointId);
-            commit.commit(checkpointId, commitMessages);
-            commitMessages.clear();
-        } catch (Exception e) {
-            throw new GeaflowRuntimeException("Failed to commit data into Paimon.", e);
+        commit(checkpointId, false);
+    }
+
+    public void commit(long checkpointId, boolean waitCompaction) {
+        flush(checkpointId, waitCompaction);
```
kitalkuyo-gita, line 87: Does flush(checkpointId, waitCompaction) ensure that commitMessages is not empty? If so, can the downstream process it?

geaflow/geaflow-plugins/geaflow-store/geaflow-store-paimon/src/main/java/org/apache/geaflow/store/paimon/commit/PaimonCommitRegistry.java
```diff
@@ -0,0 +1,80 @@
...
+        List<TaskCommitMessage> message = index2CommitMessages.computeIfAbsent(index,
+            key -> new ArrayList<>());
+        message.add(new TaskCommitMessage(tableName, commitMessages));
+    }
+
+    public synchronized List<TaskCommitMessage> pollMessages(int index) {
+        List<TaskCommitMessage> messages = index2CommitMessages.get(index);
+        if (messages != null && !messages.isEmpty()) {
+            List<TaskCommitMessage> result = new ArrayList<>(messages);
+            messages.clear();
```
kitalkuyo-gita, line 57: In concurrent scenarios, there may be data consistency risks when other threads read messages after messages.clear() is called.

geaflow/geaflow-core/geaflow-runtime/geaflow-operator/src/main/java/org/apache/geaflow/operator/impl/graph/materialize/GraphViewMaterializeOp.java
qingwen220: empty messages are filtered out before being put into the registry.

geaflow/geaflow-core/geaflow-runtime/geaflow-operator/src/main/java/org/apache/geaflow/operator/impl/graph/materialize/GraphViewMaterializeOp.java
qingwen220: no

geaflow/geaflow-core/geaflow-runtime/geaflow-operator/src/main/java/org/apache/geaflow/operator/impl/graph/materialize/PaimonGlobalSink.java
qingwen220: a task writes into its own bucket in the same table, so I think there is no need to de-duplicate and re-order messages from different tasks.  The framework will guarantees the order of messages from one task.

geaflow/geaflow-plugins/geaflow-store/geaflow-store-paimon/src/main/java/org/apache/geaflow/store/paimon/PaimonTableRWHandle.java
qingwen220: flush does not need to ensure that. It just do prepare commit.

geaflow/geaflow-core/geaflow-runtime/geaflow-operator/src/main/java/org/apache/geaflow/operator/impl/graph/materialize/PaimonGlobalSink.java
kitalkuyo-gita: > 每个任务都会写入同一张表中自己的 bucket，所以我认为没有必要对来自不同任务的消息进行去重和重新排序。框架会保证来自同一个任务的消息的顺序。

1. Although each task is responsible for its own bucket, in a distributed system, multiple tasks may concurrently write to the same table. Especially if the underlying table implementation uses mechanisms such as merging, compression, and repartitioning, message ordering and deduplication cannot rely solely on the local order of a single task.
2. The framework guarantees order within the task itself, but the final state of the global table may be affected by writes from multiple tasks. If messages are duplicated (such as in retry scenarios), the lack of deduplication can lead to data bloat or logical errors.

Regarding your question about the framework guaranteeing deduplication, Paimon explicitly requires or recommends that upstream sinks implement deduplication and ordering guarantees. Otherwise, data anomalies may occur during recovery, replay, and failover.
However, if the downstream table has a primary key constraint, duplicate messages will be discarded or overwritten. If not, data inconsistency may result.

geaflow/geaflow-plugins/geaflow-store/geaflow-store-paimon/src/main/java/org/apache/geaflow/store/paimon/BasePaimonGraphStore.java
qingwen220: Fallback is the current impl, and I currently keep the part unchanged. 
The table format string is referenced only here.

geaflow/geaflow-core/geaflow-runtime/geaflow-operator/src/main/java/org/apache/geaflow/operator/impl/graph/materialize/PaimonGlobalSink.java
kitalkuyo-gita: I'll give you a few examples to help you evaluate them.
#### Scenario 1: Task Retries Cause Data Duplicates
Suppose:

There are two tasks, TaskA (responsible for bucket 0) and TaskB (responsible for bucket 1), both writing to the same table, TableX.
TaskA wrote data D1 at checkpoint 100. Due to network jitter or a node failure, TaskA retried the write.
Data Flow:
```
TaskA first writes D1: {key: "k1", value: "v1"}
TaskA retries due to a failure and writes D1 again: {key: "k1", value: "v1"}
```
Paimon: Two identical entries appear in TableX bucket 0.
Consequences:

If the underlying table does not have a primary key constraint, multiple duplicate entries will result, leading to inaccurate query results and storage bloat.
For example, when performing count/distinct operations, the results may double.
#### Scenario 2: Out-of-order message writing affects eventual consistency
Assume:
```
Task A first writes D1 ({key: "k1", value: "v1"}) and then writes D2 ({key: "k1", value: "v2"}).
Due to distributed network latency, D2 is consumed/written to disk by downstream clients before D1 arrives.
Data flow:

After D2 is written to disk, the value of k1 changes to "v2".
D1 then writes to disk overwriting D2, and the value of k1 returns to "v1".
```
Consequence:

Without global ordering guarantees, the final table state will be inconsistent with the expected state (business-wise, k1 should be v2).
This type of "rollback" caused by out-of-order processing is common in distributed scenarios, especially during failover, recovery, and shuffle operations. 

#### Scenario 3: Data Loss or Disorganization After Partition Reassignment or Compaction
Assumptions:

Paimon internally supports bucket compaction, which may merge bucket data from multiple tasks.
TaskA writes to D1, and TaskB writes to D2. During the underlying compaction process, the data in the two buckets is merged.
Data Flow:
```
D1 and D2 are written to their own buckets.
During the compaction process, without proper deduplication and ordering logic, data in D1 and D2 may be duplicated, missing, or even overwritten out of order.
```
Consequences:

When querying TableX, multiple versions of the same key may appear, or the latest data may be lost.
"Eventual consistency" and "idempotence" cannot be guaranteed.
#### Scenario 4: Multiple Tasks Writing to the Same Bucket (Theoretical Conflict)
Assumptions:

After adjustments to certain algorithms (such as hash partitioning), TaskA and TaskB may both write to bucket 0.
Both tasks write the same key, but with different values. Data flow:
```
TaskA writes {key: "k2", value: "A"}
TaskB writes {key: "k2", value: "B"}
```
Finally, bucket 0 contains two entries with the same key but different values, and the order is uncontrollable.
Consequences:

Without global deduplication, ordering, and conflict resolution mechanisms, the final data state depends on the timing of physical writes to disk, which may not meet business expectations.

geaflow/geaflow-plugins/geaflow-store/geaflow-store-paimon/src/main/java/org/apache/geaflow/store/paimon/PaimonTableRWHandle.java
qingwen220: 1. There is only one implementation of CommitMessage.
2. added an instanceof check

geaflow/geaflow-plugins/geaflow-store/geaflow-store-paimon/src/main/java/org/apache/geaflow/store/paimon/commit/PaimonCommitRegistry.java
qingwen220: The two methods in this class are all synchronized.

geaflow/geaflow-core/geaflow-runtime/geaflow-operator/src/main/java/org/apache/geaflow/operator/impl/graph/materialize/PaimonGlobalSink.java
qingwen220: OK, add the error message in the exception.

geaflow/geaflow-core/geaflow-runtime/geaflow-operator/src/main/java/org/apache/geaflow/operator/impl/graph/materialize/PaimonGlobalSink.java
qingwen220: I don't quite understand what you said. Let's come back to the original comment. 

`For multiple messages with the same tableName, there may be merging, duplication, and order requirements, and it is necessary to ensure that they are not lost or confused.`

1. The `message` here is the paimon commit message.  Multiple messages come from different materialized tasks with each writing its own bucket. e.g. task1 writes to bucket 1, task2 writes to bucket2, .... 
2. These messages are forwarded to the downstream PaimonGlobalSink (parallelism=1), and do the final commit.

qingwen220: 1. [scenario1] iIf checkpoint 100 is successfully, Geaflow will not rewrite old data.
If checkpoint 100 is partially successful, this will not affect Paimon, because paimon commit is global and the last node of the pipeline, which means it could determine whether the whole checkpoint is successful or not. Only a successful commit could be visible to the reader. 

2. [scenario2][scenario3] Do you mean d1 and d2 are out of order between the materialized node and the global commit node? 
There are two cases: 1) d1 and d2 are from the same task,  the order is guaranteed by the shuffle module. 2) d1 and d2 are from different tasks, and they are written in different buckets. Paimon buckets are independent from each other, because they are in different LSM trees, and compaction happens just in a tree, but not across multiple trees. 

3. [scenario4] It wouldn't happen that two tasks write to the same bucket with the current implementation.

kitalkuyo-gita: OK, thank you for your answer.

geaflow/geaflow-core/geaflow-runtime/geaflow-operator/src/main/java/org/apache/geaflow/operator/impl/graph/materialize/GraphViewMaterializeOp.java
kitalkuyo-gita: ok

PR #670, opened 2025-11-14
feat: Implements lpa and cc algorithm

ChaomingZhangCN:
https://github.com/apache/geaflow/issues/360

### DIFF
--- a/geaflow/geaflow-dsl/geaflow-dsl-plan/src/main/java/org/apache/geaflow/dsl/udf/graph/ConnectedComponents.java
+++ b/geaflow/geaflow-dsl/geaflow-dsl-plan/src/main/java/org/apache/geaflow/dsl/udf/graph/ConnectedComponents.java
@@ -19,10 +19,9 @@
 
 package org.apache.geaflow.dsl.udf.graph;
 
-import java.util.ArrayList;
 import java.util.Iterator;
-import java.util.List;
 import java.util.Optional;
+import java.util.stream.Stream;
 import org.apache.geaflow.common.type.primitive.StringType;
 import org.apache.geaflow.dsl.common.algo.AlgorithmRuntimeContext;
 import org.apache.geaflow.dsl.common.algo.AlgorithmUserFunction;
@@ -62,25 +61,25 @@ public class ConnectedComponents implements AlgorithmUserFunction<Object, String
     @Override
     public void process(RowVertex vertex, Optional<Row> updatedValues, Iterator<String> messages) {
         updatedValues.ifPresent(vertex::setValue);
-        List<RowEdge> edges = new ArrayList<>(context.loadEdges(EdgeDirection.IN));
+        Stream<RowEdge> stream = context.loadEdges(EdgeDirection.IN).stream();
         if (context.getCurrentIterationId() == 1L) {
             String initValue = String.valueOf(vertex.getId());
-            sendMessageToNeighbors(edges, initValue);
+            sendMessageToNeighbors(stream, initValue);
             context.sendMessage(vertex.getId(), initValue);
             context.updateVertexValue(ObjectRow.create(initValue));
         } else if (context.getCurrentIterationId() < iteration) {
-            String minComponent = messages.next();
+            String minComponent = null;
             while (messages.hasNext()) {
                 String next = messages.next();
-                if (next.compareTo(minComponent) < 0) {
+                if (minComponent == null || next.compareTo(minComponent) < 0) {
                     minComponent = next;
                 }
             }
 
             String currentValue = (String) vertex.getValue().getField(0, StringType.INSTANCE);
             // If found smaller component id, update and propagate
-            if (minComponent.compareTo(currentValue) < 0) {
-                sendMessageToNeighbors(edges, minComponent);
+            if (minComponent != null && minComponent.compareTo(currentValue) < 0) {
+                sendMessageToNeighbors(stream, minComponent);
                 context.sendMessage(vertex.getId(), minComponent);
                 context.updateVertexValue(ObjectRow.create(minComponent));
             }
@@ -102,9 +101,7 @@ public class ConnectedComponents implements AlgorithmUserFunction<Object, String
         );
     }
 
-    private void sendMessageToNeighbors(List<RowEdge> edges, String message) {
-        for (RowEdge rowEdge : edges) {
-            context.sendMessage(rowEdge.getTargetId(), message);
-        }
+    private void sendMessageToNeighbors(Stream<RowEdge> edges, String message) {
+        edges.forEach(rowEdge -> context.sendMessage(rowEdge.getTargetId(), message));
     }
 }


geaflow/geaflow-dsl/geaflow-dsl-plan/src/main/java/org/apache/geaflow/dsl/udf/graph/ConnectedComponents.java
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    @Override
+    public void process(RowVertex vertex, Optional<Row> updatedValues, Iterator<String> messages) {
+        updatedValues.ifPresent(vertex::setValue);
+        List<RowEdge> edges = new ArrayList<>(context.loadEdges(EdgeDirection.IN));
+
+        if (context.getCurrentIterationId() == 1L) {
+            // Initialize: each vertex starts with its own id as the component id
+            String initValue = String.valueOf(vertex.getId());
```
Leomrlin, line 66: The ID column is not necessarily a string type; you can use its IType.

geaflow/geaflow-dsl/geaflow-dsl-plan/src/main/java/org/apache/geaflow/dsl/udf/graph/ConnectedComponents.java
```diff
@@ -0,0 +1,115 @@
...
+    private String keyFieldName = "component";
+    private int iteration = 1000;
+
+    @Override
+    public void init(AlgorithmRuntimeContext<Object, String> context, Object[] parameters) {
+        this.context = context;
+        if (parameters.length > 2) {
+            throw new IllegalArgumentException(
+                "Only support zero or more arguments, false arguments "
+                    + "usage: func([iteration, [keyFieldName]])");
```
Leomrlin: This can be easily misunderstood. One might first assume that keyFieldName refers to the column used to compute the cc algo. However, in reality, this parameter specifies the column name for the resulting cc values in the output. It would be helpful to add a clarification note.

geaflow/geaflow-dsl/geaflow-dsl-plan/src/main/java/org/apache/geaflow/dsl/udf/graph/ConnectedComponents.java
```diff
@@ -0,0 +1,115 @@
...
+
+        if (context.getCurrentIterationId() == 1L) {
+            // Initialize: each vertex starts with its own id as the component id
+            String initValue = String.valueOf(vertex.getId());
+            sendMessageToNeighbors(edges, initValue);
+            context.sendMessage(vertex.getId(), initValue);
+            context.updateVertexValue(ObjectRow.create(initValue));
+        } else if (context.getCurrentIterationId() < iteration) {
+            // Find the minimum component id from messages
+            String minComponent = messages.next();
```
Leomrlin: Is there definitely a message here?

geaflow/geaflow-dsl/geaflow-dsl-plan/src/main/java/org/apache/geaflow/dsl/udf/graph/ConnectedComponents.java
```diff
@@ -0,0 +1,115 @@
...
+        }
+        if (parameters.length > 1) {
+            keyFieldName = String.valueOf(parameters[1]);
+        }
+    }
+
+    @Override
+    public void process(RowVertex vertex, Optional<Row> updatedValues, Iterator<String> messages) {
+        updatedValues.ifPresent(vertex::setValue);
+        List<RowEdge> edges = new ArrayList<>(context.loadEdges(EdgeDirection.IN));
```
Leomrlin: Propagate only in the IN direction?

geaflow/geaflow-dsl/geaflow-dsl-plan/src/main/java/org/apache/geaflow/dsl/udf/graph/ConnectedComponents.java
```diff
@@ -0,0 +1,115 @@
...
+            }
+
+            String currentValue = (String) vertex.getValue().getField(0, StringType.INSTANCE);
+            // If found smaller component id, update and propagate
+            if (minComponent.compareTo(currentValue) < 0) {
+                sendMessageToNeighbors(edges, minComponent);
+                context.sendMessage(vertex.getId(), minComponent);
+                context.updateVertexValue(ObjectRow.create(minComponent));
+            } else {
+                context.sendMessage(vertex.getId(), currentValue);
```
Leomrlin: In the CC (connected components) algorithm, messages do not need to be sent when vertex values stop updating. Continuously sending messages keeps vertices active and prevents convergence, exhausting the default 1000 iteration limit and reducing efficiency.

geaflow/geaflow-dsl/geaflow-dsl-plan/src/main/java/org/apache/geaflow/dsl/udf/graph/LabelPropagation.java
```diff
@@ -0,0 +1,121 @@
...
+    private String keyFieldName = "label";
+    private int iteration = 1000;
+
+    @Override
+    public void init(AlgorithmRuntimeContext<Object, String> context, Object[] parameters) {
+        this.context = context;
+        if (parameters.length > 2) {
+            throw new IllegalArgumentException(
+                "Only support zero or more arguments, false arguments "
+                    + "usage: func([iteration, [keyFieldName]])");
```
Leomrlin: ditto

geaflow/geaflow-dsl/geaflow-dsl-plan/src/main/java/org/apache/geaflow/dsl/udf/graph/LabelPropagation.java
```diff
@@ -0,0 +1,121 @@
...
+        updatedValues.ifPresent(vertex::setValue);
+        List<RowEdge> edges = new ArrayList<>(context.loadEdges(EdgeDirection.BOTH));
+
+        if (context.getCurrentIterationId() == 1L) {
+            String initLabel = String.valueOf(vertex.getId());
+            context.updateVertexValue(ObjectRow.create(initLabel));
+            sendMessageToNeighbors(edges, initLabel);
+        } else if (context.getCurrentIterationId() < iteration) {
+            Map<String, Integer> labelCounts = new HashMap<>();
+            String currentLabel = (String) vertex.getValue().getField(0, StringType.INSTANCE);
```
Leomrlin, line 75: ditto, The ID column is not necessarily a string type; you can use its IType.

ChaomingZhangCN: The 1st iteration will update vertex value with a string type.

geaflow/geaflow-dsl/geaflow-dsl-plan/src/main/java/org/apache/geaflow/dsl/udf/graph/ConnectedComponents.java
ChaomingZhangCN: Thanks, I'll address it.

geaflow/geaflow-dsl/geaflow-dsl-plan/src/main/java/org/apache/geaflow/dsl/udf/graph/ConnectedComponents.java
ChaomingZhangCN: Compared with WeakConnectedComponents, ConnectedComponents ONLY Propagates in one direction.

geaflow/geaflow-dsl/geaflow-dsl-plan/src/main/java/org/apache/geaflow/dsl/udf/graph/LabelPropagation.java
```diff
@@ -0,0 +1,121 @@
...
+        }
+        if (parameters.length > 1) {
+            outputKeyName = String.valueOf(parameters[1]);
+        }
+    }
+
+    @Override
+    public void process(RowVertex vertex, Optional<Row> updatedValues, Iterator<String> messages) {
+        updatedValues.ifPresent(vertex::setValue);
+        List<RowEdge> edges = new ArrayList<>(context.loadEdges(EdgeDirection.BOTH));
```
kitalkuyo-gita, line 67: The way neighbor IDs are sent might be incorrect (causing spontaneous or duplicate sending across all edges).

Load edges: `context.loadEdges(EdgeDirection.BOTH)`, then iterate through the edges object `rowEdge.getTargetId()` and call `context.sendMessage(...)`.

For `BOTH`, the framework returns edges in both directions (source->target and target->source). Directly retrieving `targetId` might send the message back to itself (if `targetId == currentId`) or duplicate the same neighbor twice (if they were originally represented by each other).

Recommendation: The "other end" endpoint ID should be explicitly retrieved (see the recommendations for understanding ConnectedComponents).

geaflow/geaflow-dsl/geaflow-dsl-plan/src/main/java/org/apache/geaflow/dsl/udf/graph/LabelPropagation.java
```diff
@@ -0,0 +1,121 @@
...
+            }
+
+            String mostFrequentLabel = currentLabel;
+            int maxCount = 0;
+
+            for (Map.Entry<String, Integer> entry : labelCounts.entrySet()) {
+                String label = entry.getKey();
+                int count = entry.getValue();
+                if (count >= maxCount && label.compareTo(mostFrequentLabel) < 0) {
+                    mostFrequentLabel = label;
```
kitalkuyo-gita, line 89: After counting neighbor labels, the condition for selecting mostFrequentLabel is `if (count >= maxCount && label.compareTo(mostFrequentLabel) < 0)`.

This condition combination has two points that need review:

- 1. Using `>=` instead of `>`: When `count == maxCount`, lexicographic compare is considered, but only when `label < mostFrequentLabel` (lexicographically smaller) will a replacement be made. This biases towards selecting labels with smaller lexicographical order as tie-breaks. While the logic still works when `maxCount` is initially 0 and `mostFrequentLabel` is initially `currentLabel`, semantically, the tie-break strategy should be explicit (random, largest ID, or smallest ID). The current approach is arbitrary but deterministic.

- 2. When `labelCounts` is empty (no message received), `mostFrequentLabel` retains `currentLabel` (which is reasonable).

Recommendation:

Explicitly define and comment out the tie-break strategy. Common options:

- 1. Random selection (requires pseudo-random numbers and may affect repeatability);

- 2. Select the label with the smallest/largest id (if the label is based on the vertex id) — this is deterministic and easily interpretable;

- 3. Keep the current label (do not change it if there is no absolute majority) — this can reduce jitter.

For better repeatability and interpretability, it is recommended to use numerical comparison (if the label can be parsed into a long integer id); otherwise, use lexicographical order and comment it out as a design decision.

geaflow/geaflow-dsl/geaflow-dsl-plan/src/main/java/org/apache/geaflow/dsl/udf/graph/ConnectedComponents.java
```diff
@@ -0,0 +1,110 @@
...
+        updatedValues.ifPresent(vertex::setValue);
+        List<RowEdge> edges = new ArrayList<>(context.loadEdges(EdgeDirection.IN));
+        if (context.getCurrentIterationId() == 1L) {
+            String initValue = String.valueOf(vertex.getId());
+            sendMessageToNeighbors(edges, initValue);
+            context.sendMessage(vertex.getId(), initValue);
+            context.updateVertexValue(ObjectRow.create(initValue));
+        } else if (context.getCurrentIterationId() < iteration) {
+            String minComponent = messages.next();
+            while (messages.hasNext()) {
```
kitalkuyo-gita, line 72: Initializing the label to a string of vertex.id is common practice. Retaining the current label when messages are empty is reasonable, but the `process` method lacks protection for `messages.hasNext()`. Although the code uses `while(messages.hasNext())`, it doesn't check at the beginning (the problem is more serious with `ConnectedComponents`). `LabelPropagation` is safe here (because the `while` loop is skipped), but implementation consistency still needs attention.

geaflow/geaflow-dsl/geaflow-dsl-plan/src/main/java/org/apache/geaflow/dsl/udf/graph/ConnectedComponents.java
kitalkuyo-gita: Copying the results of `context.loadEdges(...)` to an `ArrayList(List<RowEdge> edges = new ArrayList<>(...))` can consume a lot of memory and time, especially with large datasets. Consider using the `stream` API for this purpose.

Leomrlin:
@ChaomingZhangCN This branch has conflicts that need to be resolved. Please pull the latest changes from the master branch, resolve the conflicts, and push a fix. Thx.

kaori-seasons (approved):
LGTM

Loognqiang (approved):
LGTM

Issue #677, opened 2025-11-17
feat: implement GraphSAGE algorithm based on GeaFlow inference framework


qingwen220:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

kitalkuyo-gita:
Detailed development plan to implement GraphSAGE based on GeaFlow inference framework

## 1. Framework architecture analysis

### 1.1 Core mechanism of GeaFlow-Infer inference framework

GeaFlow-Infer is a Java-Python hybrid inference framework that implements model inference through the following mechanisms:

**Core components:**
- **InferContext**: The inference context on the Java side, responsible for managing the inference task life cycle
- **Python Inference Service**: Communicates with Java processes through shared memory queues
- **DataBridge**: Implement Java-Python data exchange

### 1.2 Integration of reasoning framework and graph computing

The graph calculation passed`IncGraphInferContext`Interface call reasoning capabilities:

### 1.3 Python-side reasoning session mechanism

## 2. GraphSAGE algorithm implementation plan

### 2.1 Implement path selection

GraphSAGE can be implemented in **two ways**:

#### **Option A: Pure Infer framework implementation (recommended for inference scenarios)**
- Suitable for online inference on trained GraphSAGE models
- Utilizing geaflow-infer’s Python model calling capability
- Graph sampling is completed on the Java side, and feature aggregation is completed on the Python side through the model.

#### **Option B: UDGA framework implementation (suitable for training scenarios)**
- based on`AlgorithmUserFunction`Interface implementation
- Implement graph sampling and aggregation logic entirely on the Java side
- Suitable for scenarios where GraphSAGE training needs to be integrated into graph computing

**This solution focuses on solution A**, because it is more in line with the original design intention of geaflow-infer.

## 3. Detailed implementation plan (Plan A: based on Infer framework)

### 3.1 Overall architecture design

```mermaid
graph TB
    A[Java: IncVertexCentricCompute] --> B[邻居采样]
    B --> C[构造特征向量]
    C --> D[IncGraphInferContext.infer]
    D --> E[共享内存队列]
    E --> F[Python: GraphSAGE模型]
    F --> G[Mean/LSTM/Pool Aggregator]
    G --> H[返回节点嵌入]
    H --> E
    E --> D
    D --> I[更新节点表示]
```

### 3.2 Java side implementation

#### 3.2.1 Create GraphSAGE calculation task

[...]

PR #679, opened 2025-11-17
feat: implement GraphSAGE algorithm based on GeaFlow inference framework 

kitalkuyo-gita:
Related to [issue-677](https://github.com/apache/geaflow/issues/677)

### How was this PR tested?

### DIFF
--- a/geaflow/geaflow-core/geaflow-runtime/geaflow-operator/src/main/java/org/apache/geaflow/operator/impl/graph/compute/dynamic/DynamicGraphVertexCentricComputeOp.java
+++ b/geaflow/geaflow-core/geaflow-runtime/geaflow-operator/src/main/java/org/apache/geaflow/operator/impl/graph/compute/dynamic/DynamicGraphVertexCentricComputeOp.java
@@ -20,8 +20,10 @@
 package org.apache.geaflow.operator.impl.graph.compute.dynamic;
 
 import java.io.IOException;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import org.apache.geaflow.api.function.iterator.RichIteratorFunction;
 import org.apache.geaflow.api.graph.base.algo.AbstractIncVertexCentricComputeAlgo;
@@ -165,11 +167,21 @@ public class DynamicGraphVertexCentricComputeOp<K, VV, EV, M, FUNC extends IncVe
         public IncGraphInferComputeContextImpl() {
             if (clientLocal.get() == null) {
                 try {
-                    // Use InferContextPool instead of direct instantiation
-                    // This ensures efficient reuse of InferContext instances
-                    inferContext = InferContextPool.getOrCreate(runtimeContext.getConfiguration());
+                    // Build the effective configuration for this algorithm's InferContext.
+                    // If the algorithm declares its own Python transform class name
+                    // (code-based approach), override the global config with it so that
+                    // each algorithm can use an independent Python subprocess.
+                    // This resolves the UDF naming conflict when multiple algorithms
+                    // with different Python models run in the same job.
+                    Configuration inferConfig = buildInferConfig();
+                    inferContext = InferContextPool.getOrCreate(inferConfig);
                     clientLocal.set(inferContext);
-                    LOGGER.debug("InferContext obtained from pool: {}", 
+                    LOGGER.info(
+                        "InferContext obtained from pool for algorithm '{}', "
+                            + "pythonTransformClass='{}', pool={}",
+                        function.getClass().getSimpleName(),
+                        inferConfig.getString(
+                            FrameworkConfigKeys.INFER_ENV_USER_TRANSFORM_CLASSNAME),
                         InferContextPool.getStatus());
                 } catch (Exception e) {
                     LOGGER.error("Failed to obtain InferContext from pool", e);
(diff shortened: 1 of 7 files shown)


geaflow/geaflow-core/geaflow-runtime/geaflow-operator/src/main/java/org/apache/geaflow/operator/impl/graph/compute/dynamic/DynamicGraphVertexCentricComputeOp.java
```diff
@@ -164,11 +165,17 @@ class IncGraphInferComputeContextImpl<OUT> extends IncGraphComputeContextImpl im
...
+                    // This ensures efficient reuse of InferContext instances
+                    inferContext = InferContextPool.getOrCreate(runtimeContext.getConfiguration());
+                    clientLocal.set(inferContext);
+                    LOGGER.debug("InferContext obtained from pool: {}", 
+                        InferContextPool.getStatus());
                 } catch (Exception e) {
-                    throw new GeaflowRuntimeException(e);
+                    LOGGER.error("Failed to obtain InferContext from pool", e);
+                    throw new GeaflowRuntimeException(
+                        "InferContext initialization failed: " + e.getMessage(), e);
```
qingwen220, line 189: I don't understand the change here. clientLocal is a thread-local variable that has the same lifecycle as the graph operator.

geaflow/geaflow-infer/src/main/java/org/apache/geaflow/infer/InferContextPool.java
```diff
@@ -0,0 +1,249 @@
...
+     * This allows supporting multiple InferContext instances with different settings.
+     *
+     * @param config The configuration
+     * @return A unique key for this configuration
+     */
+    private static String generateConfigKey(Configuration config) {
+        // Use configuration hash code as the key
+        // In production, this could be enhanced with explicit key parameters
+        return "infer_" + Integer.toHexString(config.hashCode());
+    }
```
qingwen220, line 248: If the configuration is cloned, more than one infercontext will be created in a working thread.

geaflow/geaflow-infer/src/main/java/org/apache/geaflow/infer/InferDependencyManager.java
```diff
@@ -61,6 +62,10 @@ private void init() {
         }
         String pythonFilesDirectory = environmentContext.getInferFilesDirectory();
         InferFileUtils.prepareInferFilesFromJars(pythonFilesDirectory);
+        
+        // Copy user-defined UDF files (e.g., TransFormFunctionUDF.py)
+        copyUserDefinedUDFFiles(pythonFilesDirectory);
```
qingwen220, line 67: InferFileUtils.prepareInferFilesFromJars is designed to handle the download of Python UDF dependencies.

Why do we need copy again here ?

geaflow/geaflow-infer/src/main/java/org/apache/geaflow/infer/InferEnvironmentContext.java
```diff
@@ -63,14 +64,53 @@ public class InferEnvironmentContext {
...
+        this.virtualEnvDirectory = virtualEnvDirectory != null ? virtualEnvDirectory : "";
         this.inferFilesDirectory = pythonFilesDirectory;
-        this.inferLibPath = virtualEnvDirectory + LIB_PATH;
-        this.pythonExec = virtualEnvDirectory + PYTHON_EXEC;
-        this.inferScript = pythonFilesDirectory + INFER_SCRIPT_FILE;
         this.roleNameIndex = queryRoleNameIndex();
         this.configuration = configuration;
         this.envFinished = false;
+        
+        // Check if using system Python
```
qingwen220, line 73: GeaFlow infer framework use the miniconda as the python env. 
You can refer to the configuration FrameworkConfigKeys.INFER_ENV_CONDA_URL.

geaflow/geaflow-infer/src/main/resources/infer/inferRuntime/SPSCQueueWrite.h
```diff
@@ -60,7 +60,7 @@ class SPSCQueueWrite : public SPSCQueueBase
   public:
     SPSCQueueWrite(const char* fileName, int64_t len): SPSCQueueBase(mmap(fileName, len), len), toMove_(0) {}
 
-    ~SPSCQueueWrite() {}
+    virtual ~SPSCQueueWrite() {}
```
qingwen220, line 64: Why do we need virtual here?

geaflow/geaflow-infer/src/main/resources/infer/inferRuntime/mmap_ipc.pyx
```diff
@@ -28,8 +28,8 @@ from libc.stdint cimport *
 cdef extern from "MmapIPC.h":
     cdef cppclass MmapIPC:
         MmapIPC(char* , char*) except +
-        int readBytes(int) nogil except +
-        bool writeBytes(char *, int) nogil except +
```
qingwen220, line 32: why do we need the change?

geaflow/geaflow-dsl/geaflow-dsl-plan/src/main/resources/TransFormFunctionUDF.py
```diff
@@ -0,0 +1,534 @@
...
+            neighbor_features: List of tensors, each of shape [in_dim]
+            
+        Returns:
+            Aggregated feature tensor of shape [out_dim]
+        """
+        if len(neighbor_features) == 0:
+            # No neighbors, just apply linear transformation to node features
+            output = self.linear_without_neighbors(node_feature)
+        else:
+            # Stack neighbors: [num_neighbors, in_dim]
```
qingwen220, line 517: I was thinking that it might be better to implement GraphSAGE as the example algorithm, similar to PageRank. Since putting Python functions into the engine's JAR package could easily cause conflicts with users' own UDFs, this approach would be safer.

kitalkuyo-gita: Hello @qingwen220 . I believe we should consider incorporating Python UDFs, as a significant number of models are developed in Python, and the ecosystem of libraries supporting Python-based models is exceptionally rich. While the Java example functions should certainly be retained, I consider Python UDFs to be equally important. The specific code sections you commented on above are designed to isolate the relevant UDFs within a local Python virtual environment (Venv) during the development phase, thereby requiring the user to specify a custom path for their UDFs.

geaflow/geaflow-infer/src/main/java/org/apache/geaflow/infer/InferDependencyManager.java
kitalkuyo-gita: This is to make it easier for users to debug when developing UDF functions.

geaflow/geaflow-infer/src/main/java/org/apache/geaflow/infer/InferContextPool.java
kitalkuyo-gita: This is for reusing and creating the inferContext within the inferpool, thus serving as a resource reuse mechanism.

geaflow/geaflow-core/geaflow-runtime/geaflow-operator/src/main/java/org/apache/geaflow/operator/impl/graph/compute/dynamic/DynamicGraphVertexCentricComputeOp.java
kitalkuyo-gita: The change itself was made because it was feared that the current inference context (inferContext) could not be obtained from inferpool. clientLocal was made to use the already created inferContext, reusing existing resources.

geaflow/geaflow-dsl/geaflow-dsl-plan/src/main/resources/TransFormFunctionUDF.py
qingwen220: That's what I mean. My preference is to expose it as a Python UDF.

PR #693, opened 2025-11-30
Adding parser changes for OPTIONAL MATCH keyword

vamossagar12:
Adding parser changes for OPTIONAL MATCH keyword. It seems it is part of the ISO GQL standard and is an equivalent of outer join in SQL.

### How was this PR tested?


vamossagar12:
#694

PR #716, opened 2025-12-17
feat(ai): Adding Lucene & Embedding-Based Search Operators to Apache GeaFlow (incubating) for Lightweight Context Memory

Leomrlin:
We're excited to introduce initial support for **context-aware memory operations** in Apache GeaFlow (incubating) through the integration of two key retrieval operators: **Lucene-powered keyword search** and **embedding-based semantic search**. This enhancement lays the foundational layer for building dynamic, AI-driven graph memory systems — enabling real-time, hybrid querying over structured graph data and unstructured semantic intent.

### ✅ Key Features Implemented

- **`KeywordVector` + Lucene Indexing**: Enables fast, full-text retrieval of entities using BM25-style keyword matching. Ideal for surfacing exact or near-exact matches from entity attributes (e.g., names, emails, titles).
- **`EmbeddingVector` + Vector Index Store**: Supports semantic search via high-dimensional embeddings. Queries are encoded using a configured embedding model and matched against pre-indexed node representations.
- **Hybrid `VectorSearch` Interface**: Combines multiple vector types (keyword, embedding, traversal hints) into a single search context, paving the way for multimodal retrieval.
- **End-to-End Query Pipeline**: From query ingestion → hybrid indexing → graph retrieval → context verbalization, demonstrated with LDBC-scale data.

### 🧪 Validated Use Cases

Our `GraphMemoryTest` suite demonstrates:
- Resolving ambiguous queries like _"Chaim Azriel"_ into multiple candidate persons using **keyword + embedding fusion**.
- Traversing relationships (e.g., `Comment_hasCreator_Person`) in follow-up rounds via **contextual refinement**.
- Iterative context expansion across multiple search cycles — mimicking agent memory evolution.

### 🔮 Why This Matters

This work represents the first step toward **Graphiti-inspired, relationship-aware AI memory** within GeaFlow:
> Instead of treating context as static text, we model it as a **dynamic, evolving subgraph**, enriched by both semantic similarity and topological structure.

By leveraging GeaFlow’s native streaming graph engine, we aim to go beyond batch RAG — supporting **incremental updates**, **temporal reasoning**, and **multi-hop inference** at low latency.

---

**Next Steps**:  
We propose incubating this as the **GeaFlow Memory Engine**, with upcoming support for:
- Graph traversal-guided re-ranking
- Agent session management with episodic memory
- Integration with LLM agents for autonomous reasoning

This PR sets the stage: **from graph analytics to graph-native AI memory**.

Let’s build the future of contextual intelligence — on streaming graphs. 🚀

### DIFF
--- a/geaflow-ai/src/main/java/org/apache/geaflow/ai/operator/SearchUtils.java
+++ b/geaflow-ai/src/main/java/org/apache/geaflow/ai/operator/SearchUtils.java
@@ -42,7 +42,7 @@ public class SearchUtils {
      * @return an unmodifiable set of ignored characters
      */
     private static Set<Character> buildIgnoredChars() {
-        Set<Character> ignored = new HashSet<>(EXCLUDED_CHARS);
+        Set<Character> ignored = new HashSet<>(32);
         // Add digits
         for (char c = '0'; c <= '9'; c++) {
             ignored.add(c);
@@ -98,7 +98,7 @@ public class SearchUtils {
             return false; // Consider empty/null invalid; adjust based on use case
         }
         for (char c : str.toCharArray()) {
-            if (!IGNORE_CHARS.contains(c)) {
+            if (IGNORE_CHARS.contains(c)) {
                 return false;
             }
         }
--- a/geaflow-ai/src/main/java/org/apache/geaflow/ai/operator/SessionOperator.java
+++ b/geaflow-ai/src/main/java/org/apache/geaflow/ai/operator/SessionOperator.java
@@ -81,6 +81,7 @@ public class SessionOperator implements SearchOperator {
             }
             //recall compute
             GraphSearchStore searchStore = initSearchStore(extendEntityIndexMap);
+            searchStore.close();
             List<GraphEntity> matchEntities = searchStore.search(query, graphAccessor);
             Set<GraphEntity> matchEntitiesSet = new HashSet<>(matchEntities);
 
@@ -122,6 +123,7 @@ public class SessionOperator implements SearchOperator {
         }
         //recall compute
         GraphSearchStore searchStore = initSearchStore(entityIndexMap);
+        searchStore.close();
         return searchStore.search(query, graphAccessor);
     }
 
@@ -136,7 +138,6 @@ public class SessionOperator implements SearchOperator {
                 }
             }
         }
-        searchStore.close();
         return searchStore;
     }
 }
--- a/geaflow-ai/src/main/java/org/apache/geaflow/ai/session/SessionManagement.java
+++ b/geaflow-ai/src/main/java/org/apache/geaflow/ai/session/SessionManagement.java
@@ -19,10 +19,7 @@
 
 package org.apache.geaflow.ai.session;
 
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.UUID;
+import java.util.*;
 import org.apache.geaflow.ai.common.config.Constants;
 import org.apache.geaflow.ai.subgraph.SubGraph;
 
@@ -40,17 +37,14 @@ public class SessionManagement {
             return false;
         }
         session2ActiveTime.put(sessionId, System.nanoTime());
+        session2Graphs.putIfAbsent(sessionId, new ArrayList<>());
         return true;
     }
 
     public String createSession() {
         String sessionId = Constants.PREFIX_TMP_SESSION + System.nanoTime()
                 + UUID.randomUUID().toString().replace("-", "").substring(0, 8);
-        if (createSession(sessionId)) {
-            return sessionId;
-        } else {
-            return null;
-        }
+        return createSession(sessionId) ? sessionId : null;
     }
 
     public boolean sessionExists(String session) {
@@ -58,7 +52,8 @@ public class SessionManagement {
     }
 
     public List<SubGraph> getSubGraph(String sessionId) {
-        return this.session2Graphs.get(sessionId);
+        List<SubGraph> l = this.session2Graphs.get(sessionId);
+        return l == null ? new ArrayList<>() : l;
     }
 
     public void setSubGraph(String sessionId, List<SubGraph> subGraphs) {
--- a/geaflow-ai/src/main/java/org/apache/geaflow/ai/verbalization/SubgraphSemanticPromptFunction.java
+++ b/geaflow-ai/src/main/java/org/apache/geaflow/ai/verbalization/SubgraphSemanticPromptFunction.java
@@ -72,12 +72,12 @@ public class SubgraphSemanticPromptFunction implements VerbalizationFunction {
         if (entity instanceof GraphVertex) {
             GraphVertex graphVertex = (GraphVertex) entity;
             return graphVertex.getVertex().getValues().stream()
-                    .filter(str -> !SearchUtils.isAllAllowedChars(str))
+                    .filter(SearchUtils::isAllAllowedChars)
                     .map(SearchUtils::formatQuery).collect(Collectors.toList());
         } else if (entity instanceof GraphEdge) {
             GraphEdge graphEdge = (GraphEdge) entity;
             return graphEdge.getEdge().getValues().stream()
-                    .filter(str -> !SearchUtils.isAllAllowedChars(str))
+                    .filter(SearchUtils::isAllAllowedChars)
                     .map(SearchUtils::formatQuery).collect(Collectors.toList());
         }
         return new ArrayList<>();


geaflow-ai/pom.xml
```diff
@@ -0,0 +1,127 @@
...
+        <version>0.8.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>geaflow-ai</artifactId>
+
+    <properties>
+        <solon-mcp.version>3.5.1</solon-mcp.version>
+        <junit.version>5.10.1</junit.version>
+        <slf4j.version>1.7.15</slf4j.version>
+        <log4j.version>1.2.17</log4j.version>
```
Appointat: test

geaflow-ai/src/main/java/org/apache/geaflow/ai/common/model/EmbeddingService.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    public String singleSentence(String sentence) {
+        OfflineModelDirect model = new OfflineModelDirect();
+        ModelContext context = ModelContext.emptyContext();
+        context.setModelInfo(modelInfo);
+        context.userSay(sentence);
+        return model.chat(context);
+    }
+
```
Appointat, line 35: The method name singleSentence() is not accurate; it actually sends a single message and retrieves a reply.
Suggestion: rename to chat() or sendMessage()

geaflow-ai/src/main/java/org/apache/geaflow/ai/common/model/EmbeddingService.java
```diff
@@ -0,0 +1,84 @@
...
+        List<EmbeddingResult> embeddingResults = model.embedding(context);
+        Gson gson = new Gson();
+        StringBuilder builder = new StringBuilder();
+        for (EmbeddingResult result : embeddingResults) {
+            builder.append("\n");
+            String json = gson.toJson(result);
+            builder.append(json);
+        }
+        return builder.toString();
+    }
```
Appointat, line 48: ChatRobot is responsible for both chat and embedding, with a mixed set of responsibilities.

Suggestion: consider splitting into ChatService and EmbeddingService? May be better?

geaflow-ai/src/main/java/org/apache/geaflow/ai/common/model/EmbeddingService.java
```diff
@@ -0,0 +1,84 @@
...
+
+    public static class EmbeddingResult {
+        public String input;
+        public double[] embedding;
+
+        public EmbeddingResult(String input, double[] embedding) {
+            this.input = input;
+            this.embedding = embedding;
+        }
```
Appointat, line 57: what about merge it to "EmbeddingResponse"?

geaflow-ai/src/main/java/org/apache/geaflow/ai/common/model/ModelInfo.java
```diff
@@ -0,0 +1,70 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.geaflow.ai.common.model;
+
+public class ModelInfo {
```
Appointat: Rename to ModelConfig may be better. This is the mainstream naming.

geaflow-ai/src/main/java/org/apache/geaflow/ai/common/model/Response.java
```diff
@@ -0,0 +1,65 @@
...
+ * under the License.
+ */
+
+package org.apache.geaflow.ai.common.model;
+
+import com.google.gson.annotations.SerializedName;
+import java.util.List;
+
+
+public class Response {
```
Appointat, line 26: If you need to integrate an API response compatible with the OpenAI/Gemini API, please import these variables.  Viewing the confusing point in this code, why is the `choice` needed? If the `Response` class is a generic class (not only for LLM responses), then I think usage and choice may be not necessary; these information could be stored as meta attributes of the message.

geaflow-ai/src/test/resources/graph_ldbc_sf/Comment/part-00000-0ea3b18b-e932-4946-9be9-d17b750eafea-c0001.csv
```diff
@@ -0,0 +1,58 @@
+creationDate|id|locationIP|browserUsed|content|length
```
Appointat, line 1: Can you upload these CSV data somewhere else, instead of the `src`? Or can you package these data into some cloud/remote storage? Otherwise, this data would make the Java package larger. (However, given that the current PR is still very preliminary, I think this comment can be ignored.)

geaflow-ai/src/main/java/org/apache/geaflow/ai/common/model/OfflineModelDirect.java
```diff
@@ -0,0 +1,70 @@
...
+ */
+
+package org.apache.geaflow.ai.common.model;
+
+import com.google.gson.Gson;
+import java.util.ArrayList;
+import java.util.List;
+import org.apache.geaflow.common.utils.RetryCommand;
+
+public class OfflineModelDirect {
```
Appointat: I have some misunderstandings about the term 'offline'. In the context, does 'offline' mean the “offline (batch) processing/calling” model (like LLM) relative to streaming computation? Perhaps OnlineModelClient or RemoteModelClient would be more accurate.

geaflow-ai/src/main/java/org/apache/geaflow/ai/common/model/RemoteModelClient.java
```diff
@@ -0,0 +1,70 @@
...
+            }
+
+        }
+        return null;
+    }
+
+    public List<ChatRobot.EmbeddingResult> embedding(ModelEmbedding context) {
+        ModelInfo info = context.getModelInfo();
+        OkHttpDirectConnector connector = new OkHttpDirectConnector(
+                info.getUrl(), info.getApi(), info.getUserToken());
```
Appointat, line 52: Creating a new OkHttpClient instance on every call leads to: establishing a new connection each time and resource leaks. 
Since we’re not yet at a level to optimize this, it is recommended to add a TODO: reuse the Connector or OkHttpClient as a singleton or as a member variable.

geaflow-ai/src/main/java/org/apache/geaflow/ai/common/model/OfflineModelDirect.java
```diff
@@ -0,0 +1,70 @@
...
+
+
+    public String chat(ModelContext context) {
+        ModelInfo info = context.getModelInfo();
+        OkHttpDirectConnector connector = new OkHttpDirectConnector(
+                info.getUrl(), info.getApi(), info.getUserToken());
+        String request = new Gson().toJson(context);
+        org.apache.geaflow.ai.common.model.Response response = connector.post(request);
```
Appointat: embedding() has a retry mechanism (10 attempts, 3-second interval) but chat() does not retry. Suggestion: uniformly add a retry mechanism.

geaflow-ai/src/main/java/org/apache/geaflow/ai/common/model/RemoteModelClient.java
```diff
@@ -0,0 +1,70 @@
...
+            return new ArrayList<>();
+        }
+        List<ChatRobot.EmbeddingResult> embeddingResults = new ArrayList<>();
+        for (EmbeddingResponse.EmbeddingVector v : response.data) {
+            int index = v.index;
+            String input = context.input[index];
+            double[] vector = v.embedding;
+            ChatRobot.EmbeddingResult result = new ChatRobot.EmbeddingResult(input, vector);
+            embeddingResults.add(result);
+        }
```
Appointat, line 75: response.data may be null. No check of context.input[index]. If the index is out of bounds, an exception will be thrown.

geaflow-ai/src/main/java/org/apache/geaflow/ai/common/model/OkHttpDirectConnector.java
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    public org.apache.geaflow.ai.common.model.Response post(String bodyJson) {
+        RequestBody requestBody = RequestBody.create(
+                MediaType.parse("application/json; charset=utf-8"),
+                bodyJson
+        );
+
+        String url = endpoint + useApi;
+        System.out.println(url);
```
Appointat: Please use a logger. Other similar code also needs to be corrected.

geaflow-ai/src/main/java/org/apache/geaflow/ai/common/model/OkHttpDirectConnector.java
```diff
@@ -0,0 +1,112 @@
...
+        try (okhttp3.Response response = client.newCall(request).execute()) {
+            if (response.isSuccessful() && response.body() != null) {
+                String responseBody = response.body().string();
+                return GSON.fromJson(responseBody, org.apache.geaflow.ai.common.model.Response.class);
+            } else {
+                System.out.println("Request failed with code: " + response.code());
+            }
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
```
Appointat, line 84: The logic for catching function exceptions can be optimized:

- post() catches the exception, prints the stack trace, and returns null;
- embeddingPost() wraps it in a RuntimeException and throws it.

Therefore, a unified approach is recommended.

geaflow-ai/src/main/java/org/apache/geaflow/ai/graph/LocalMemoryGraphAccessor.java
```diff
@@ -0,0 +1,152 @@
...
+
+    @Override
+    public GraphVertex getVertex(String label, String id) {
+        return new GraphVertex(graph.getVertex(label, id));
+    }
+
+    @Override
+    public GraphEdge getEdge(String label, String src, String dst) {
+        return new GraphEdge(graph.getEdge(label, src, dst));
+    }
```
Appointat, line 70: getVertex/getEdge may return wrapped objects that contain null, so a non-null check is required.

Appointat: The other parts of the code also have similar risks and need to be checked one by one.

geaflow-ai/src/main/java/org/apache/geaflow/ai/graph/LocalFileGraphAccessor.java
```diff
@@ -0,0 +1,152 @@
...
+        public GraphVertex next() {
+            if (!hasNext()) {
+                throw new NoSuchElementException();
+            }
+            Vertex nextVertex = vertexIterator.next();
+            return new GraphVertex(nextVertex);
+        }
+
+        @Override
+        public void remove() {
```
Appointat: It is a read-only graph accessor, why does it support remove? (This comment may be related to personal development habits; I generally think the accessor is read-only)

geaflow-ai/src/main/java/org/apache/geaflow/ai/graph/LocalFileGraphAccessor.java
```diff
@@ -0,0 +1,152 @@
...
+        }
+
+        @Override
+        public boolean hasNext() {
+            return vertexIterator.hasNext();
+        }
+
+        @Override
+        public GraphVertex next() {
+            if (!hasNext()) {
```
Appointat: If I’m not mistaken, vertexIterator.next() itself checks and throws NoSuchElementException. Therefore, GraphVertexIterator.next() performs a duplicate hasNext check. It is recommended to delegate this exception checking to the underlying iterator.

geaflow-ai/src/main/java/org/apache/geaflow/ai/graph/GraphVertex.java
```diff
@@ -0,0 +1,63 @@
...
+ */
+
+package org.apache.geaflow.ai.graph;
+
+import java.util.Objects;
+import org.apache.geaflow.ai.graph.io.Vertex;
+
+public class GraphVertex implements GraphEntity {
+
+    public final Vertex vertex;
```
Appointat: Be `private`?

geaflow-ai/src/main/java/org/apache/geaflow/ai/index/vector/KeywordVector.java
```diff
@@ -0,0 +1,65 @@
...
+        this.vec = vec;
+    }
+
+    public String[] getVec() {
+        return vec;
+    }
+
+    @Override
+    public double match(IVector other) {
+        if (other.getType() != this.getType()) {
```
Appointat: I hope that the way to determine the type inside this is consistent (in general, getType is more efficient, but the code will be a bit dirtier/redundant). Reference code:

```
// EmbeddingVector - use instanceof
if (!(other instanceof EmbeddingVector)) {
    return 0.0;
}

// KeywordVector - use getType()
if (other.getType() != this.getType()) {
    return 0.0;
}
```

geaflow-ai/src/main/java/org/apache/geaflow/ai/index/vector/KeywordVector.java
```diff
@@ -0,0 +1,65 @@
...
+        KeywordVector otherKeyword = (KeywordVector) other;
+        int count = 0;
+        for (String keyword1 : this.vec) {
+            for (String keyword2 : otherKeyword.vec) {
+                if (keyword1.equals(keyword2)) {
+                    count++;
+                    break;
+                }
+            }
+        }
```
Appointat, line 55: add a TODO: use HashSet to optimaize it to O(n)

geaflow-ai/src/main/java/org/apache/geaflow/ai/index/vector/TraversalVector.java
```diff
@@ -0,0 +1,65 @@
...
+
+package org.apache.geaflow.ai.index.vector;
+
+public class TraversalVector implements IVector {
+
+    private final String[] vec;
+
+    public TraversalVector(String... vec) {
+        if (vec.length % 3 != 0) {
+            throw new RuntimeException("Traversal vector shold be src-edge-dst pairs");
```
Appointat: shold -> should

geaflow-ai/src/main/java/org/apache/geaflow/ai/index/vector/TraversalVector.java
```diff
@@ -0,0 +1,65 @@
...
+
+    @Override
+    public VectorType getType() {
+        return VectorType.TraversalVector;
+    }
+
+    @Override
+    public String toString() {
+
+        StringBuilder builder = new StringBuilder();
```
Appointat:
```
@Override
public String toString() {
    StringBuilder sb = new StringBuilder("TraversalVector{vec=");
    
    for (int i = 0; i < vec.length; i++) {
        if (i > 0) {
            sb.append(i % 3 == 0 ? "; " : "-");
        }
        sb.append(vec[i]);
        if (i % 3 == 2) {
            sb.append(">");
        }
    }
    
    return sb.append('}').toString();
}
```

geaflow-ai/src/main/java/org/apache/geaflow/ai/index/EmbeddingIndexStore.java
```diff
@@ -0,0 +1,251 @@
...
+
+        System.out.println("Success to read index store file. items num: " + count);
+        System.out.println("Success to rebuild index with file. index num: " + this.indexStoreMap.size());
+
+
+        //Scan entities in the graph, make new index items
+        ChatRobot chatRobot = new ChatRobot();
+        chatRobot.setModelInfo(modelInfo);
+
+        final int BATCH_SIZE = 32;
```
Appointat: Magic numbers should be extracted as constants or configured or environment variables. Other magic numbers should also be changed.

geaflow-ai/src/main/java/org/apache/geaflow/ai/graph/io/CsvFileReader.java
```diff
@@ -0,0 +1,149 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
```
Appointat, line 2: I have not reviewed the io folder yet.

geaflow-ai/src/main/java/org/apache/geaflow/ai/operator/SessionOperator.java
```diff
@@ -0,0 +1,142 @@
...
+            entityIndexMap.put(vertex, vertexIndex);
+        }
+        //recall compute
+        GraphSearchStore searchStore = initSearchStore(entityIndexMap);
+        return searchStore.search(query, graphAccessor);
+    }
+
+    private GraphSearchStore initSearchStore(Map<GraphEntity, List<IVector>> entityIndexMap) {
+        GraphSearchStore searchStore = new GraphSearchStore();
+        for (Map.Entry<GraphEntity, List<IVector>> entry : entityIndexMap.entrySet()) {
```
Appointat, line 132: Do we rebuild the in-memory index on every search (GraphSearchStore includes Lucene)? Is there a better optimization method? For example, clearing/initializing the GraphSearchStore instead of rebuilding.

geaflow-ai/src/main/java/org/apache/geaflow/ai/operator/EmbeddingOperator.java
```diff
@@ -0,0 +1,193 @@
...
+
+    private final GraphAccessor graphAccessor;
+    private final IndexStore indexStore;
+    private double threshold;
+    private int topN;
+
+    public EmbeddingOperator(GraphAccessor accessor, IndexStore store) {
+        this.graphAccessor = Objects.requireNonNull(accessor);
+        this.indexStore = Objects.requireNonNull(store);
+        this.threshold = 0.50;
```
Appointat: Should it be placed in the configuration, or, another optiona -- as a configurable parameter?

Appointat (changes requested):
Thanks for your PR. Left some comments.

geaflow-ai/src/main/java/org/apache/geaflow/ai/common/model/EmbeddingService.java
Leomrlin: `chat()` has been replaced here

geaflow-ai/src/main/java/org/apache/geaflow/ai/common/model/EmbeddingService.java
Leomrlin: The two functions have been separate into `ChatService` and `EmbeddingService`.

geaflow-ai/src/main/java/org/apache/geaflow/ai/common/model/EmbeddingService.java
Leomrlin: `EmbeddingResponse` and `EmbeddingResult` have no common fields and serve different purposes, so they are not suitable for merging

geaflow-ai/src/main/java/org/apache/geaflow/ai/common/model/ModelInfo.java
Leomrlin: Yes, it's now renamed to `ModelConfig`.

geaflow-ai/src/main/java/org/apache/geaflow/ai/common/model/Response.java
Leomrlin: `choice` is part of the interface and must be retained. Currently, `Response` is not yet a generic class, so it only converts the model's response output.

geaflow-ai/src/test/resources/graph_ldbc_sf/Comment/part-00000-0ea3b18b-e932-4946-9be9-d17b750eafea-c0001.csv
Leomrlin: These CSV files have been specially processed to keep them within 100 lines, making them suitable as test data for unit tests.

geaflow-ai/src/main/java/org/apache/geaflow/ai/common/model/OfflineModelDirect.java
Leomrlin: OK, it has now been renamed to `RemoteModelClient`.

geaflow-ai/src/main/java/org/apache/geaflow/ai/common/model/RemoteModelClient.java
Leomrlin: A shared client is now added here, avoiding to create a new client each time.

geaflow-ai/src/main/java/org/apache/geaflow/ai/common/model/OfflineModelDirect.java
Leomrlin: Added retries

geaflow-ai/src/main/java/org/apache/geaflow/ai/common/model/RemoteModelClient.java
Leomrlin: Added null checks.

geaflow-ai/src/main/java/org/apache/geaflow/ai/common/model/OkHttpDirectConnector.java
Leomrlin: Added LOGGERs to classes that require output logging. All outputs now use LOGGER.

geaflow-ai/src/main/java/org/apache/geaflow/ai/common/model/OkHttpDirectConnector.java
Leomrlin: Added log output and threw a `RuntimeException` here

geaflow-ai/src/main/java/org/apache/geaflow/ai/graph/LocalMemoryGraphAccessor.java
Leomrlin: There was confusion here.
Now we consistently use `null` to represent no result, and if the value is non-null, the internal entities are also guaranteed to be non-null. Null checks have been added at the usage sites.

geaflow-ai/src/main/java/org/apache/geaflow/ai/graph/LocalFileGraphAccessor.java
Leomrlin: The graph accessor is not suitable for exposing an iterator with a remove interface. Here, the `remove()` method has been removed and replaced with an exception throw.

geaflow-ai/src/main/java/org/apache/geaflow/ai/graph/LocalFileGraphAccessor.java
Leomrlin: Removed redundant checks.

geaflow-ai/src/main/java/org/apache/geaflow/ai/graph/GraphVertex.java
Leomrlin: Already modified.

geaflow-ai/src/main/java/org/apache/geaflow/ai/index/vector/KeywordVector.java
Leomrlin: For simplicity, `instanceof` is used here for all type checks.

geaflow-ai/src/main/java/org/apache/geaflow/ai/index/vector/KeywordVector.java
Leomrlin: Modified the implementation here to reduce its time complexity to O(n).

geaflow-ai/src/main/java/org/apache/geaflow/ai/index/vector/TraversalVector.java
Leomrlin: modified

geaflow-ai/src/main/java/org/apache/geaflow/ai/index/vector/TraversalVector.java
Leomrlin: It has been modified accordingly.

geaflow-ai/src/main/java/org/apache/geaflow/ai/index/EmbeddingIndexStore.java
Leomrlin: Currently, `model.Constants` and `GraphMemoryConfigKeys` have been added, and all necessary literal values have been moved and declared within them.

geaflow-ai/src/main/java/org/apache/geaflow/ai/graph/io/CsvFileReader.java
Leomrlin: It's okay.

geaflow-ai/src/main/java/org/apache/geaflow/ai/operator/SessionOperator.java
Leomrlin: Unfortunately, there is no better solution at the moment. The algorithm requires that during each iteration, the `Operator` searches within the potential traversal area. Using the previous `GraphSearchStore`—whether based on the previous iteration or the entire graph—would inevitably lead to an expanding search scope. Fortunately, though, the retrieval area is not particularly large in the graph.

geaflow-ai/src/main/java/org/apache/geaflow/ai/operator/EmbeddingOperator.java
Leomrlin: ditto, already modified.

geaflow-ai/src/main/java/org/apache/geaflow/ai/session/SessionManagement.java
```diff
@@ -0,0 +1,67 @@
...
+    public boolean sessionExists(String session) {
+        return this.session2ActiveTime.containsKey(session);
+    }
+
+    public List<SubGraph> getSubGraph(String sessionId) {
+        return this.session2Graphs.get(sessionId);
+    }
+
+    public void setSubGraph(String sessionId, List<SubGraph> subGraphs) {
+        this.session2Graphs.put(sessionId, subGraphs);
```
kitalkuyo-gita, line 60: `SessionManagement.createSession(String)` only writes the time to `session2ActiveTime` but does not create the corresponding empty list in `session2Graphs`. `GraphMemoryServer.verbalize` directly calls `sessionManagement.getSubGraph(sessionId)`; if `subGraphList` is null when calling `new ArrayList<>(subGraphList.size())`, it will throw a NullPointerException. 

It is recommended to call `session2Graphs.put(sessionId, new ArrayList<>())` when `createSession(String)` and `createSession()` are successful.
Also, change `getSubGraph` to return a non-null value.

kitalkuyo-gita: samples:
```
// 将 Map 改为并发实现
private final ConcurrentMap<String, Long> session2ActiveTime = new ConcurrentHashMap<>();
private final ConcurrentMap<String, List<SubGraph>> session2Graphs = new ConcurrentHashMap<>();

public boolean createSession(String sessionId) {
    if (sessionId == null) {
        return false;
    }
    Long prev = session2ActiveTime.putIfAbsent(sessionId, System.nanoTime());
    if (prev != null) {
        return false;
    }
    // 初始化 subgraphs 为可变空列表，避免 NPE
    session2Graphs.putIfAbsent(sessionId, new ArrayList<>());
    return true;
}

public String createSession() {
    String sessionId = Constants.PREFIX_TMP_SESSION + System.nanoTime()
            + UUID.randomUUID().toString().replace("-", "").substring(0, 8);
    return createSession(sessionId) ? sessionId : null;
}

// 返回不可为 null 的 List（防止调用者 NPE）
public List<SubGraph> getSubGraph(String sessionId) {
    List<SubGraph> l = this.session2Graphs.get(sessionId);
    return l == null ? new ArrayList<>() : l;
}

public void setSubGraph(String sessionId, List<SubGraph> subGraphs) {
    // 安全性：确保 map 存在 key
    this.session2Graphs.put(sessionId, subGraphs == null ? new ArrayList<>() : subGraphs);
}
```

geaflow-ai/src/main/java/org/apache/geaflow/ai/operator/SearchUtils.java
```diff
@@ -0,0 +1,108 @@
...
+    private static final Set<Character> IGNORE_CHARS = buildIgnoredChars();
+
+    /**
+     * Builds the set of allowed characters for input validation.
+     * Includes alphanumeric characters and selected common symbols.
+     *
+     * @return an unmodifiable set of ignored characters
+     */
+    private static Set<Character> buildIgnoredChars() {
+        Set<Character> ignored = new HashSet<>(EXCLUDED_CHARS);
```
kitalkuyo-gita: It is recommended not to include EXCLUDED_CHARS in IGNORE_CHARS, as this may cause errors in SearchStore query string construction and semantic filtering.

For example, `SubgraphSemanticPromptFunction.verbalize` filters strings using `.filter(str -> !SearchUtils.isAllAllowedChars(str))`. An incorrect set of allowed characters will lead to incorrect filtering behavior (strings that should be kept are removed, and strings that should be removed are kept).

samples:

```
// SearchUtils.java: 修复 buildIgnoredChars()
private static final Set<Character> EXCLUDED_CHARS = new HashSet<>(Arrays.asList(
        '*', '#', '-', '?', '`', '{', '}', '[', ']', '(', ')', '>', '<', ':', '/', '.'
));

private static final Set<Character> IGNORE_CHARS = buildIgnoredChars();

private static Set<Character> buildIgnoredChars() {
    Set<Character> allowed = new HashSet<>();
    // 加入英文字母（大小写）
    for (char c = 'a'; c <= 'z'; c++) allowed.add(c);
    for (char c = 'A'; c <= 'Z'; c++) allowed.add(c);
    // 加入数字
    for (char c = '0'; c <= '9'; c++) allowed.add(c);
    // 加入常用安全字符（空格、下划线等）
    allowed.add(' ');
    allowed.add('_');
    allowed.add('-');
    allowed.add('@');
    allowed.add('+');
    allowed.add('!');
    allowed.add('$');
    allowed.add('%');
    allowed.add('&');
    allowed.add('=');
    allowed.add('~');
    // 不要加入 EXCLUDED_CHARS !
    return Collections.unmodifiableSet(allowed);
}

```

geaflow-ai/src/main/java/org/apache/geaflow/ai/operator/SessionOperator.java
```diff
@@ -0,0 +1,142 @@
...
+            //Traverse all extension points of the subgraph and search within the extension area
+            for (SubGraph subGraph : subGraphList) {
+                List<GraphEntity> extendEntities = getSubgraphExpand(subGraph);
+                for (GraphEntity extendEntity : extendEntities) {
+                    List<IVector> entityIndex = indexStore.getEntityIndex(extendEntity);
+                    extendEntityIndexMap.put(extendEntity, entityIndex);
+                }
+            }
+            //recall compute
+            GraphSearchStore searchStore = initSearchStore(extendEntityIndexMap);
```
kitalkuyo-gita, line 83: Perhaps it would be better to explicitly call `writer.commit()` (or `close()`) in `initSearchStore` after all the `addDoc` operations are completed?

Leomrlin: Here it has been changed to explicitly call `close()`.

geaflow-ai/src/main/java/org/apache/geaflow/ai/operator/SearchUtils.java
Leomrlin: Removed `EXCLUDED_CHARS` from `IGNORE_CHARS`

geaflow-ai/src/main/java/org/apache/geaflow/ai/session/SessionManagement.java
Leomrlin: An empty `SubGraph` list is created in `session2Graphs` when a session is created.

DukeWangYu:
LGTM

kitalkuyo-gita (approved):
LGTM

Appointat (approved):
LGTM

PR #729, opened 2026-01-21
feat(ai): support graph consolidate algorithm

Leomrlin:
### Description
The current context memory system lacks support for unstructured data types such as text, images, video, and audio. These data formats do not come with pre-built graph structures, requiring a mechanism to dynamically extract and model relationships from raw data.

### Solution
Implement a `ConsolidateFunction` component that will:
1. Read raw unstructured data from the storage system
2. Leverage available computational resources to:
   - Analyze content and extract semantic relationships
   - Generate necessary association edges between data entities
   - Create searchable indices for different data modalities
3. Model the generated metadata as nodes/edges in the graph structure
4. Write the processed graph data back to the storage system
5. Support both initial graph construction and incremental updates

### DIFF
--- a/geaflow-ai/src/main/java/org/apache/geaflow/ai/GeaFlowMemoryServer.java
+++ b/geaflow-ai/src/main/java/org/apache/geaflow/ai/GeaFlowMemoryServer.java
@@ -160,6 +160,8 @@ public class GeaFlowMemoryServer {
                 memoryMutableGraph.addEdge(((GraphEdge) entity).getEdge());
             }
         }
+        CACHE.getConsolidateServer().executeConsolidateTask(
+            CACHE.getServerByName(graphName).getGraphAccessors().get(0), memoryMutableGraph);
         return "Success to add entities, num: " + graphEntities.size();
     }
 
--- a/geaflow-ai/src/main/java/org/apache/geaflow/ai/service/ServerMemoryCache.java
+++ b/geaflow-ai/src/main/java/org/apache/geaflow/ai/service/ServerMemoryCache.java
@@ -23,6 +23,7 @@ import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.Map;
 import org.apache.geaflow.ai.GraphMemoryServer;
+import org.apache.geaflow.ai.consolidate.ConsolidateServer;
 import org.apache.geaflow.ai.graph.Graph;
 
 public class ServerMemoryCache {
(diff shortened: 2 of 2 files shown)


geaflow-ai/src/main/java/org/apache/geaflow/ai/consolidate/function/KeywordRelationFunction.java
```diff
@@ -0,0 +1,102 @@
...
+        indexStore.initStore(new SubgraphSemanticPromptFunction(graphAccessor));
+        LOGGER.info("Success to init EntityAttributeIndexStore.");
+        GraphMemoryServer server = new GraphMemoryServer();
+        server.addGraphAccessor(graphAccessor);
+        server.addIndexStore(indexStore);
+        LOGGER.info("Success to init GraphMemoryServer.");
+
+        if (null == mutableGraph.getSchema().getSchema(
+            Constants.CONSOLIDATE_KEYWORD_RELATION_LABEL)) {
+            int code = mutableGraph.addEdgeSchema(new EdgeSchema(
```
cbqiao, line 54: If the graph schema is changed or updated inside a function rather than being created by the caller before execution, is there a risk that the updated schema will not be visible to subsequent steps?

geaflow-ai/src/main/java/org/apache/geaflow/ai/GeaFlowMemoryServer.java
```diff
@@ -0,0 +1,230 @@
...
+import org.apache.geaflow.ai.service.ServerMemoryCache;
+import org.apache.geaflow.ai.verbalization.Context;
+import org.apache.geaflow.ai.verbalization.SubgraphSemanticPromptFunction;
+import org.noear.solon.Solon;
+import org.noear.solon.annotation.*;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@Controller
+public class GeaFlowMemoryServer {
```
cbqiao, line 41: Is there a need for an API to start the ConsolidateServer? I don't see any explicit startup command for it.

geaflow-ai/src/main/java/org/apache/geaflow/ai/consolidate/function/KeywordRelationFunction.java
Leomrlin: The modification and synchronization of the mutableGraph are entirely managed by the external HTAP graph engine. The Server side of the GraphMemory component is not responsible for this logic. Currently, there is only one instance of ConsolidateServer in the system, so there are no data synchronization or conflict issues between multiple instances.

geaflow-ai/src/main/java/org/apache/geaflow/ai/GeaFlowMemoryServer.java
Leomrlin: Added a common `ConsolidateServer` in the `CACHE`

cbqiao:
LGTM

PR #749, opened 2026-02-28
fix: update RocksdbClient to use getDeclaredConstructor for options instantiation 

SYaoJun:
update some the deprecated methods to use new methods.
1. update RocksdbClient to use getDeclaredConstructor for options instantiation
2. replace deprecated KnnVectorField with KnnFloatVectorField in GraphVectorIndex

```shell
The method newInstance() from the type Class<capture#1-of ?> is deprecated
The type KnnVectorField is deprecatedJava(16777221)
```

### How was this PR tested?

### DIFF
--- a/geaflow/geaflow-plugins/geaflow-store/geaflow-store-rocksdb/src/main/java/org/apache/geaflow/store/rocksdb/RocksdbClient.java
+++ b/geaflow/geaflow-plugins/geaflow-store/geaflow-store-rocksdb/src/main/java/org/apache/geaflow/store/rocksdb/RocksdbClient.java
@@ -88,7 +88,7 @@ public class RocksdbClient {
         if (this.rocksDBOptions == null || this.rocksDBOptions.isClosed()) {
             LOGGER.info("rocksdb optionClass {}", optionClass);
             try {
-                this.rocksDBOptions = (IRocksDBOptions) Class.forName(optionClass).newInstance();
+                this.rocksDBOptions = (IRocksDBOptions) Class.forName(optionClass).getDeclaredConstructor().newInstance();
                 this.rocksDBOptions.init(config);
             } catch (Throwable e) {
                 LOGGER.error("{} not found", optionClass);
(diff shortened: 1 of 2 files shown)


geaflow/geaflow-plugins/geaflow-store/geaflow-store-rocksdb/src/main/java/org/apache/geaflow/store/rocksdb/RocksdbClient.java
```diff
@@ -88,7 +88,7 @@ private void initRocksDbOptions() {
         if (this.rocksDBOptions == null || this.rocksDBOptions.isClosed()) {
             LOGGER.info("rocksdb optionClass {}", optionClass);
             try {
-                this.rocksDBOptions = (IRocksDBOptions) Class.forName(optionClass).newInstance();
+                this.rocksDBOptions = (IRocksDBOptions) Class.forName(optionClass).getDeclaredConstructor().newInstance();
```
Leomrlin, line 91: Do all implementation classes of `optionClass` have a no-argument constructor? 
If some classes do not have a no-argument constructor, will `getDeclaredConstructor()` throw an exception?

geaflow/geaflow-plugins/geaflow-store/geaflow-store-vector/src/main/java/org/apache/geaflow/store/lucene/GraphVectorIndex.java
```diff
@@ -142,12 +142,12 @@ public K searchVectorIndex(boolean isVertex, String fieldName, float[] vector, i
...
 
             // Create KNN vector query
-            KnnVectorQuery knnQuery = new KnnVectorQuery(fieldName, vector, topK);
+            KnnFloatVectorQuery knnQuery = new KnnFloatVectorQuery(fieldName, vector, topK);
 
             // Execute search
             TopDocs topDocs = searcher.search(knnQuery, topK);
 
-            Document firstDoc = searcher.doc(topDocs.scoreDocs[0].doc);
+            Document firstDoc = searcher.storedFields().document(topDocs.scoreDocs[0].doc);
```
Leomrlin, line 150: What are the version requirements for `Lucene` regarding the `storedFields()` method? Does this change affect performance (by adding an extra method call)?

PR #750, opened 2026-02-28
fix: improve error handling in searchVectorIndex method

SYaoJun:
ensure IndexReader is closed properly

### How was this PR tested?

### DIFF
--- a/geaflow/geaflow-plugins/geaflow-store/geaflow-store-vector/src/main/java/org/apache/geaflow/store/lucene/GraphVectorIndex.java
+++ b/geaflow/geaflow-plugins/geaflow-store/geaflow-store-vector/src/main/java/org/apache/geaflow/store/lucene/GraphVectorIndex.java
@@ -136,10 +136,7 @@ public class GraphVectorIndex<K> implements IVectorIndex<K> {
      */
     @Override
     public K searchVectorIndex(boolean isVertex, String fieldName, float[] vector, int topK) {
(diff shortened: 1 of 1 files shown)


geaflow/geaflow-plugins/geaflow-store/geaflow-store-vector/src/main/java/org/apache/geaflow/store/lucene/GraphVectorIndex.java
```diff
@@ -166,11 +167,17 @@ public K searchVectorIndex(boolean isVertex, String fieldName, float[] vector, i
...
-
             return result;
         } catch (IOException e) {
             throw new RuntimeException("Failed to search vector index", e);
+        } finally {
+            if (reader != null) {
+                try {
+                    reader.close();
+                } catch (IOException e) {
+                    throw new RuntimeException("Failed to close IndexReader", e);
```
Leomrlin: With this change, the exception from `reader.close()` could mask the original exception.

SYaoJun: > With this change, the exception from `reader.close()` could mask the original exception.

Thanks for the suggestions. Refactored resource management to use try-with-resources, eliminating the risk of suppressed exceptions during IndexReader closure.

Issue #758, opened 2026-03-03
Math.abs(Integer.MIN_VALUE) will overflow and return a negative value


SYaoJun:
**Describe the bug**

[Substr.java](https://github.com/apache/geaflow/blob/main/geaflow/geaflow-dsl/geaflow-dsl-plan/src/main/java/org/apache/geaflow/dsl/udf/table/string/Substr.java#L34-L36)

```java
if ((Math.abs(pos) > str.length())) {
    return str;
}
```
**Expected behavior**

Math.abs(Integer.MIN_VALUE) will overflow
return a negative value

PR #759, opened 2026-03-03
fix(udf): handle Integer.MIN_VALUE in Substr function

SYaoJun:

fix: #758 

### How was this PR tested?

### DIFF
--- a/geaflow/geaflow-dsl/geaflow-dsl-plan/src/main/java/org/apache/geaflow/dsl/udf/table/string/Substr.java
+++ b/geaflow/geaflow-dsl/geaflow-dsl-plan/src/main/java/org/apache/geaflow/dsl/udf/table/string/Substr.java
@@ -31,6 +31,9 @@ public class Substr extends UDF {
         if (str == null || pos == null || length == null) {
             return null;
         }
+        if (pos == Integer.MIN_VALUE) {
+            return str;
+        }
         if ((Math.abs(pos) > str.length())) {
             return str;
         }
@@ -62,6 +65,9 @@ public class Substr extends UDF {
 
     public BinaryString eval(BinaryString str, Integer pos, Integer length) {
         if (str == null || pos == null || length == null) {
+            return null;
+        }
+        if (pos == Integer.MIN_VALUE) {
             return null;
         }
         if (Math.abs(pos) > str.getLength()) {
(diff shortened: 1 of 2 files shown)


geaflow/geaflow-dsl/geaflow-dsl-plan/src/test/java/org/apache/geaflow/dsl/udf/string/SubstrTest.java
```diff
@@ -43,4 +43,17 @@ public void test() {
         Assert.assertEquals(sb.eval(BinaryString.fromString("Facebook"), -5).toString(), "ebook");
         Assert.assertEquals(sb.eval(BinaryString.fromString("Facebook"), 5, 1).toString(), "b");
     }
+
+    @Test
+    public void testIntegerMinValue() {
+        Substr sb = new Substr();
+
+        // Test Integer.MIN_VALUE for String version
+        Assert.assertEquals(sb.eval("hello", Integer.MIN_VALUE, 1), "hello");
```
Leomrlin, line 52: Is this usage not common in SQL?

geaflow/geaflow-dsl/geaflow-dsl-plan/src/main/java/org/apache/geaflow/dsl/udf/table/string/Substr.java
```diff
@@ -64,6 +67,9 @@ public BinaryString eval(BinaryString str, Integer pos, Integer length) {
         if (str == null || pos == null || length == null) {
             return null;
         }
+        if (pos == Integer.MIN_VALUE) {
+            return null;
```
Leomrlin, line 71: Are the String and BinaryString versions inconsistent?

SYaoJun: Yes, the String and BinaryString versions are inconsistent. More precisely, the return value of those if conditiona are based on below three lines.

[in String](https://github.com/apache/geaflow/pull/759/changes#diff-2089f75b29ae56fc4e0e7d5d70b9cb08a1f3b56dbced9281aceeab42d5fd302aL34)
```java
if ((Math.abs(pos) > str.length())) {
    return str;
}
```
[in BinaryString](https://github.com/apache/geaflow/pull/759/changes#diff-2089f75b29ae56fc4e0e7d5d70b9cb08a1f3b56dbced9281aceeab42d5fd302aL68)
```java
if (Math.abs(pos) > str.getLength()) {
    return null;
}
```

PR #760, opened 2026-03-03
fix(math): handle null and empty inputs in average methods

SYaoJun:

Add tests when array and list is empty to avoid throw devide zero exception.
### How was this PR tested?


Leomrlin (approved):
LGTM

PR #769, opened 2026-03-06
fix:Add check for empty search results to prevent array out of bounds

SYaoJun:

Directly accessing index 0 carries a risk.
It is better to check that the length is greater than zero in advance.

```java
Document firstDoc = searcher.doc(topDocs.scoreDocs[0].doc);
```

### How was this PR tested?


Leomrlin:
It looks like this PR might be conflicting with the one your just merged recently. Could you take a look and fix it when you have a moment? @SYaoJun

PR #781, opened 2026-03-14
Implement MagnitudeVector and add corresponding test cases

### LINKED ISSUE #773: [feat] MagnitudeVector and TraversalVector not implementation
I noticed that these two classes still haven't implemented the match methods. I'm interested in having a go at implementing them—could you provide some detailed explanations about these two classes?

SYaoJun:

fix: #773 

add MagnitudeVector with match logic and unit tests

### How was this PR tested?

### DIFF
--- a/geaflow-ai/src/main/java/org/apache/geaflow/ai/index/vector/MagnitudeVector.java
+++ b/geaflow-ai/src/main/java/org/apache/geaflow/ai/index/vector/MagnitudeVector.java
@@ -19,20 +19,80 @@
 
 package org.apache.geaflow.ai.index.vector;
 
+import java.util.Objects;
+
 public class MagnitudeVector implements IVector {
+
+    private final double magnitude;
+
+    public MagnitudeVector() {
+        this.magnitude = 0.0;
+    }
+
+    public MagnitudeVector(double magnitude) {
+        this.magnitude = magnitude;
+    }
+
+    public double getMagnitude() {
+        return magnitude;
+    }
 
     @Override
     public double match(IVector other) {
-        return 0;
+        if (!(other instanceof MagnitudeVector)) {
+            throw new IllegalArgumentException("Other vector must be a MagnitudeVector");
+        }
+
+        MagnitudeVector otherVec = (MagnitudeVector) other;
+        double otherMagnitude = otherVec.magnitude;
+
+        return computeSimilarity(otherMagnitude);
+       
     }
 
+    private double computeSimilarity(double otherMagnitude) {
+        if (this.magnitude == 0.0 && otherMagnitude == 0.0) {
+            return 1.0;
+        }
+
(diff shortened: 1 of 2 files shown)


SYaoJun:
@Leomrlin Could you please take a look at this code when you have a moment?

geaflow-ai/src/main/java/org/apache/geaflow/ai/index/vector/MagnitudeVector.java
```diff
@@ -19,20 +19,76 @@
...
+    }
+
+    public double getMagnitude() {
+        return magnitude;
+    }
+
     @Override
     public double match(IVector other) {
-        return 0;
+        if (!(other instanceof MagnitudeVector)) {
```
kitalkuyo-gita, line 45: Code snippet: `if (!(other instanceof MagnitudeVector)) { throw new IllegalArgumentException("Other vector must be a MagnitudeVector"); }`

Problem: This can cause runtime exceptions when calling `match` without explicitly performing type checking when combining multiple `IVector`s (EmbeddingVector, KeywordVector, TraversalVector, etc.) in `VectorSearch`, potentially interrupting the search process.

Recommendation: For `IVector`s of different types, it's advisable to return 0.0 (indicating dissimilarity/not included in the scoring), or have a more explicit contract defined by the `IVector` interface (e.g., `match` supports different types and returns 0). Returning 0 is more robust and consistent with the design comment in `paste` ("if not same type return 0.0").

geaflow-ai/src/main/java/org/apache/geaflow/ai/index/vector/MagnitudeVector.java
```diff
@@ -19,20 +19,76 @@
...
+
+        MagnitudeVector otherVec = (MagnitudeVector) other;
+        double otherMagnitude = otherVec.magnitude;
+
+        return computeSimilarity(otherMagnitude);
+       
+    }
+
+    private double computeSimilarity(double otherMagnitude) {
+        if (this.magnitude == 0.0 && otherMagnitude == 0.0) {
```
kitalkuyo-gita, line 54: In the code, "magnitude" typically refers to metrics such as degree or PageRank—values ​​that are generally non-negative (≥0)—though certain centrality measures (e.g., specific eigenvector-based or normalized metrics) can be negative. The current implementation handles this using `Math.abs`, ensuring the calculation proceeds; however, it is advisable to either explicitly define the valid range (≥0) in the class documentation or formally document the meaning of negative values ​​if they are permitted.

Recommendation: Add Javadoc to the constructor specifying the expected range, or implement validation within the constructor (throwing an `IllegalArgumentException` if only non-negative values ​​are allowed).

SYaoJun: Good suggestions. I have updated the code. Please take a look again!

PR #810, opened 2026-06-29
Add a ClickHouse connector with batched writes

### LINKED ISSUE #798: Add a ClickHouse connector (efficient batch writes)
### What & why

ClickHouse is a popular OLAP database, often the downstream analytics layer for
graph-computing results. You can technically connect today via the JDBC
connector, but it lacks ClickHouse-tuned batch writes (buffering + columnar
writes), so performance suffers.

ClickHouse 是流行的 OLAP 数据库,常作为图计算结果的下游分析层。现在用 JDBC connector
勉强能连,但缺针对 ClickHouse 的高效批量写(攒批 + 列式写入),性能差。

### The task

Create `geaflow-dsl-connector-clickhouse`, focusing the sink on efficient batched
writes (buffer + flush threshold); the source should support parallel partitioned reads.

新建 `geaflow-dsl-connector-clickhouse`,重点优化 sink 批量写(buffer + flush 阈值),
source 支持按分区并行读。

### Where to look / 怎么做

1. Use `geaflow-dsl-connector-jdbc` as a baseline and spot the differences
   between "generic JDBC" and "ClickHouse-specific".
2. Sink batching: `write()` goes into a buffer; commit in bulk on threshold or `flush()`.
3. SPI registration + parent/aggregation pom + docs + tests.

### Done when

- [ ] Sink throughput is clearly better than row-by-row JDBC (attach a simple benchmark in the PR)
- [ ] Source supports parallel partitioned reads
- [ ] Testcontainers integration test
- [ ] CN + EN docs + examples
- [ ] checkstyle / RAT pass

GabrielBBaldez:
## What

Adds `geaflow-dsl-connector-clickhouse`, a ClickHouse connector with a **batched sink** and a **partitioned source**, wired in through the connector SPI and modelled on `geaflow-dsl-connector-jdbc`.

Fixes #798

## Sink — batched writes

Rows are buffered and flushed in bulk via `PreparedStatement.addBatch()/executeBatch()` once `geaflow.dsl.clickhouse.write.batch.size` (default 1000) is reached, and again when the window finishes. ClickHouse is columnar and optimized for bulk ingestion, so this is far faster than the row-by-row inserts you get when connecting through the generic JDBC connector.

## Source — parallel partitioned reads

The read is split over a numeric column into `partition.num` ranges (the same partitioning model as the JDBC connector) so the partitions can be read in parallel.

## Tested

Testcontainers integration test against a real ClickHouse 24.3 instance:

- batched write + row-count verification, partitioned read round-trip, partition splitting, and SPI discovery (`type='clickhouse'` resolving to the connector);
- a write benchmark — batched vs. row-by-row → roughly **30x** higher throughput (logged by the test).

checkstyle and RAT pass. Chinese and English docs added under `docs/docs-cn` / `docs/docs-en` and wired into the connector `index.rst`.


geaflow/geaflow-dsl/geaflow-dsl-connector/geaflow-dsl-connector-clickhouse/src/main/java/org/apache/geaflow/dsl/connector/clickhouse/ClickHouseTableSource.java
```diff
@@ -0,0 +1,280 @@
...
+
+    @Override
+    public List<Partition> listPartitions() {
+        if (partitionNum <= 0) {
+            throw new GeaFlowDSLException("invalid clickhouse partition number: {}", partitionNum);
+        }
+        if (partitionNum == 1) {
+            return Collections.singletonList(new ClickHousePartition(tableName, ""));
+        }
+        long stride = upperBound / partitionNum - lowerBound / partitionNum;
```
kitalkuyo-gita: Incorrect partition stride calculation — potential for overlapping partitions or missed reads

Code snippet: `ClickHouseTableSource.listPartitions()`

#### Issue

The current code—`long stride = upperBound / partitionNum - lowerBound / partitionNum;`—splits the division into two separate integer divisions before subtracting, resulting in rounding errors; semantically, it should be `(upperBound - lowerBound) / partitionNum`.

#### Risk

Incorrect partition boundary calculations can lead to rows being read multiple times or not at all (due to erroneous range splitting).

#### Solution (fix and handle small/zero ranges):

Use `long range = upperBound - lowerBound; stride = Math.max(1, range / partitionNum);`
Ensure `partitionNum` does not exceed `range` (if `range < partitionNum`, cap it at `range` or `1`), and ensure all rows are covered in edge cases (with the last partition absorbing the remainder). #### Suggested code (replacing the original stride/loop section):

```
long range = upperBound - lowerBound;
if (range <= 0) {
throw new GeaFlowDSLException("invalid partition range: lowerbound=%d upperbound=%d", lowerBound, upperBound);
}
if (partitionNum > range) {
partitionNum = range; // at most one id per partition
}
long baseStride = range / partitionNum;
long remainder = range % partitionNum;
long current = lowerBound;
List<Partition> partitions = new ArrayList<>();
for (long i = 0; i < partitionNum; ++i) {
long thisStride = baseStride + (i < remainder ? 1 : 0);
long next = current + thisStride;
String lBound = (i != 0) ? String.format("%s >= %d", partitionColumn, current) : null;
String uBound = (i != partitionNum - 1) ? String.format("%s < %d", partitionColumn, next) : null;
String whereClause;
if (uBound == null) {
whereClause = lBound;
} else if (lBound == null) {
whereClause = String.format("%s OR %s IS NULL", uBound, partitionColumn);
} else {
whereClause = String.format("%s AND %s", lBound, uBound);
}
partitions.add(new ClickHousePartition(tableName, "WHERE " + whereClause));
current = next;
}
```
Explanation: Here, the remainder is distributed among the first few partitions, ensuring the entire range is fully covered without overlap.

geaflow/geaflow-dsl/geaflow-dsl-connector/geaflow-dsl-connector-clickhouse/src/main/java/org/apache/geaflow/dsl/connector/clickhouse/util/ClickHouseUtils.java
```diff
@@ -0,0 +1,111 @@
...
+
+public class ClickHouseUtils {
+
+    /**
+     * Builds a parameterized batch-insert statement of the form
+     * {@code INSERT INTO table (c1, c2, ...) VALUES (?, ?, ...)}. The same prepared statement is
+     * reused across the whole batch, which is what lets ClickHouse ingest rows in bulk instead of
+     * one round-trip per row.
+     */
+    public static String buildInsertSql(String tableName, List<TableField> fields) {
```
kitalkuyo-gita, line 45: SQL Construction and Identifier/Parameter Security — Risks of SQL Injection and Identifier Injection

Code references: `ClickHouseUtils.buildInsertSql()`, `ClickHouseUtils.selectRowsFromTable()`, `ClickHouseTableSource.listPartitions()` (involving the construction of `whereClause` and the passing of table/column names).

#### Issue

Table and column names are directly concatenated into the SQL string without validation or quoting. If a user injects a malicious string (or one containing spaces or special characters) via the configuration, it can lead to syntax errors or injection vulnerabilities.
`selectRowsFromTable` uses `Statement` combined with string concatenation for clauses such as `LIMIT`, `OFFSET`, and `ORDER BY`.

#### Risk

If the configuration source—even if generally trusted—unexpectedly contains special characters, it could cause query failures or security issues. The general rule is that all external inputs, including configurations, must be validated or escaped.

#### Solution (Two Parts)

A) Validate table and column names against an allowlist and enclose them in identifier quotes (ClickHouse supports backticks or double quotes; choose the style consistent with the project) to reject invalid identifiers. B) Switch to using `PreparedStatement` for `LIMIT` and `OFFSET` values ​​(passing parameterizable numbers as parameters) and avoid concatenating variable strings into the `whereClause` whenever possible. If the `whereClause` is constructed internally by the framework (as is currently the case), ensure that only numeric constants are inserted, or use placeholders and set the corresponding parameters. Suggested universal tool method (put in ClickHouseUtils):
```
private static final Pattern IDENTIFIER = Pattern.compile("^[A-Za-z_][A-Za-z0-9_]*$");

public static String quoteIdentifier(String id) { 
if (id == null || !IDENTIFIER.matcher(id).matches()) { 
throw new GeaFlowDSLException("invalid identifier: " + id); 
} 
return "`" + id + "`"; // ClickHouse accepts backticks
}
Modify buildInsertSql to use quoteIdentifier:
Java
public static String buildInsertSql(String tableName, List<TableField> fields) { 
String qTable = quoteIdentifier(tableName); 
String columns = fields.stream() 
.map(f -> quoteIdentifier(f.getName())) 
.collect(Collectors.joining(", ")); 
String placeholders = fields.stream().map(f -> "?").collect(Collectors.joining(", ")); 
return String.format("INSERT INTO %s (%s) VALUES (%s)", qTable, columns, placeholders);
}
```

Modify selectRowsFromTable to use PreparedStatement (and use quoteIdentifier to verify the table name/column name):

```
public static List<Row> selectRowsFromTable(Connection conn, String tableName, 
String whereClause, int columnNum, 
long startOffset, long windowSize, String orderByColumnName) 
throws SQLException { 
if (windowSize == Windows.SIZE_OF_ALL_WINDOW) { 
windowSize = Integer.MAX_VALUE; 
} else if (windowSize <= 0) { 
throw new GeaFlowDSLException("wrong windowSize"); 
} 

String qTable = quoteIdentifier(tableName); 
String qOrder = quoteIdentifier(orderByColumnName); 
// whereClause is either empty or starts with "WHERE ...". If it was constructed here (numbers), 
// it's safe. Otherwise validate/deny. 
String sql = String.format("SELECT * FROM %s %s ORDER BY %s LIMIT ? OFFSET ?", qTable, whereClause == null ? "" : whereClause, qOrder); 
try (PreparedStatement ps = conn.prepareStatement(sql)) { 
ps.setLong(1, windowSize);
ps.setLong(2, startOffset);
try (ResultSet rs = ps.executeQuery()) {
List<Row> rowList = new ArrayList<>();
while (rs.next()) {
Object[] values ​​= new Object[columnNum];
for (int i = 1; i <= columnNum; i++) {
values[i - 1] = rs.getObject(i);
}
rowList.add(ObjectRow.create(values));
}
return rowList;
}
}
}
```
Note: `PreparedStatement` cannot be used to substitute table or column names, so `quoteIdentifier` validation is crucial.

geaflow/geaflow-dsl/geaflow-dsl-connector/geaflow-dsl-connector-clickhouse/src/main/java/org/apache/geaflow/dsl/connector/clickhouse/ClickHouseTableSink.java
```diff
@@ -0,0 +1,144 @@
...
+        } catch (SQLException e) {
+            throw new GeaFlowDSLException("failed to flush batch to table: " + tableName, e);
+        }
+    }
+
+    private void flush() throws SQLException {
+        if (bufferedRows == 0) {
+            return;
+        }
+        statement.executeBatch();
```
kitalkuyo-gita: Code reference: `ClickHouseTableSink.flush()` — `executeBatch` may throw an `SQLException`; `bufferedRows` is not rolled back, and `statement.clearBatch()` is not executed when an exception occurs.

- `ClickHouseTableSink.close()`: If `close` throws an exception, it masks exceptions thrown upstream (e.g., errors during `finish`), preventing tests or users from seeing the original error.
- `ClickHouseTableSource.fetch()`: Uses `Statement`/`Connection`; specifically, `selectRowsFromTable` calls `statement.executeQuery(selectQuery)` and closes the `ResultSet` in a utility method, whereas `PreparedStatement` usage is best handled via a local try-with-resources block.

Risks: Resource leaks (open connections/statements), unclear error logs, loss of buffered row count state following batch failures, and difficulties with program retries or recovery. Suggested code modifications (flush & close):

```
private void flush() throws SQLException {
if (bufferedRows == 0) {
return;
}
try {
statement.executeBatch();
// If ClickHouse driver requires commit, optionally:
// if (!connection.getAutoCommit()) connection.commit();
LOGGER.info("flushed {} rows to clickhouse table {}", bufferedRows, tableName);
} finally {
try {
statement.clearBatch();
} catch (SQLException e) {
LOGGER.warn("failed to clear batch", e);
}
bufferedRows = 0;
}
}

@Override
public void close() {
SQLException closeEx = null;
try {
if (this.statement != null) {
try {
this.statement.close();
} catch (SQLException e) {
LOGGER.warn("failed to close statement", e);
closeEx = e;
} finally {
this.statement = null;
}
}
if (this.connection != null) {
try {
this.connection.close();
} catch (SQLException e) {
LOGGER.warn("failed to close connection", e);
if (closeEx == null) closeEx = e;
} finally {
this.connection = null;
}
}
LOGGER.info("close clickhouse sink");
} finally {
if (closeEx != null) {
// Do not swallow—wrap in a runtime exception to surface it to the caller, or log it; ensure previous exceptions are not lost.
throw new GeaFlowDSLException("failed to close clickhouse sink", closeEx);
}
}
}
```
Explanation: The example above attempts to capture exceptions encountered while closing each resource and ultimately throws them as a single exception (or, depending on project conventions, logs them as WARN and suppresses them).

kitalkuyo-gita:
@GabrielBBaldez  Hello! I’ve been extremely busy lately and have just seen your PR. Thank you very much for your valuable contribution to the community. We take a rigorous and serious approach with every new contributor; I have carefully reviewed your code and identified some performance issues, so I hope you can take a moment to look at them. Best regards.

kitalkuyo-gita (approved):
LGTM

geaflow/geaflow-dsl/geaflow-dsl-connector/geaflow-dsl-connector-clickhouse/src/main/java/org/apache/geaflow/dsl/connector/clickhouse/ClickHouseTableSource.java
GabrielBBaldez: Thank you for catching this one — you were right, and the arithmetic was wrong in a way that would have been painful to debug in production.

Dividing each bound before subtracting rounds both terms independently, so the stride drifted from the real range. It now splits the range as a whole:

```java
long stride = (upperBound - lowerBound) / partitionNum;
```

I left a comment above it explaining why the two forms are not equivalent, so nobody quietly reverts it later. Fixed in `d3f9150`.

geaflow/geaflow-dsl/geaflow-dsl-connector/geaflow-dsl-connector-clickhouse/src/main/java/org/apache/geaflow/dsl/connector/clickhouse/ClickHouseTableSink.java
GabrielBBaldez: Thank you — this was a real one. A failed `executeBatch` left the rows in the buffer, so the next flush, or `close()`, would resend and recount them.

`flush()` now drops the batch in a `finally`, so it happens whether or not the execute succeeded:

```java
} finally {
    bufferedRows = 0;
    try {
        statement.clearBatch();
    } catch (SQLException e) {
        LOGGER.warn("failed to clear batch after flush", e);
    }
}
```

`clearBatch()` is wrapped on its own so a failure there cannot mask the original `SQLException` on its way out. `close()` was hardened in the same commit (`d3f9150`), and `selectRowsFromTable` moved to try-with-resources while I was in there — the `ResultSet` leaked on any exception between `executeQuery` and the explicit `close()`.

geaflow/geaflow-dsl/geaflow-dsl-connector/geaflow-dsl-connector-clickhouse/src/main/java/org/apache/geaflow/dsl/connector/clickhouse/util/ClickHouseUtils.java
GabrielBBaldez: Thank you for raising this, and I owe you a straight answer rather than a quiet one: this is the only point of the three I did not fully address, and I should have said so at the time instead of leaving the thread open.

Where it stands today:

- **Values** are bound through `PreparedStatement` placeholders — `buildInsertSql` emits `?` per field and `bindRow` sets them by type, so row data never reaches the SQL text.
- **Identifiers** are still interpolated. `buildInsertSql` and `selectRowsFromTable` build the statement with `String.format`, so the table and column names go in unquoted and unvalidated.

In practice those names arrive from the user's own DDL rather than from the data path, so the exposure is narrower than it looks — but the failure mode you are pointing at is real, and a table or column whose name needs quoting would break the statement outright before any security question arises.

I would rather fix it than leave it noted. Two options, and I am happy either way:

1. Quote identifiers on the way in — ClickHouse accepts backticks, so wrapping each name and rejecting any that contains one keeps it simple.
2. Validate against a strict pattern and fail fast with a clear message when a name falls outside it.

Say which you prefer and I will push it here. If you would rather this PR stay as reviewed and land as it is, a follow-up issue is fine too — your call, and thank you again for the care you put into this review.

Loognqiang (approved):
LGTM

PR #811, opened 2026-07-08
[ISSUE-789] Add string functions TRIM, LPAD, RPAD

### LINKED ISSUE #789: Add the string functions `TRIM` / `LPAD` / `RPAD`
### What & why / 来龙去脉

GeaFlow's DSL already ships 27 string functions (`LTrim`, `RTrim`, `Substr`,
`Concat`, …), but three everyday ones are missing: `TRIM` (strip whitespace from
both ends), and `LPAD` / `RPAD` (pad a string to a given length). These come up
constantly in data cleaning, so adding them is an immediately useful first PR.

DSL 已经有 27 个字符串函数,但还缺三个数据清洗里天天用的:`TRIM`(去首尾空格)、
`LPAD`/`RPAD`(左右填充到指定长度)。补上它们,第一个 PR 就能让人立刻用上。

### The task / 要做的事

Add three scalar functions that match standard SQL behavior:
- `trim(str)` — strip leading/trailing whitespace
- `lpad(str, len, pad)` / `rpad(str, len, pad)` — pad to a target length

### Where to look / 改哪些文件

1. Copy the shape of
   `geaflow-dsl/geaflow-dsl-plan/src/main/java/org/apache/geaflow/dsl/udf/table/string/LTrim.java`.
   In the same folder add `Trim.java`, `LPad.java`, `RPad.java` — extend `UDF`,
   add the `@Description` annotation, implement `eval(...)`, and mind the null case.
2. Register three lines in `BuildInSqlFunctionTable.java`:
   `.add(GeaFlowFunction.of(Trim.class))`.
3. Add `function_trim.sql` (etc.) under `query/` and matching `.txt` files under `expect/`.

### Done when / 完成标准

- [ ] All three functions implemented and registered, each with `@Description`
- [ ] `null` input returns `null` rather than throwing
- [ ] New `.sql` tests + `.txt` expectations pass via `mvn test` on the right test class
- [ ] Behavior matches MySQL / standard SQL — paste a quick comparison in the PR

candy972023:
## What does this PR do?

Closes #789

Adds three string scalar functions to GeaFlow DSL:
- `trim(str)` — strip leading/trailing whitespace
- `lpad(str, len, pad)` — left-pad to target length
- `rpad(str, len, pad)` — right-pad to target length

## Key design decisions

- All functions return null on null input (no exception)
- LPad/RPad handle edge cases: len <= 0 returns empty, empty pad returns original
- Trim supports both String and BinaryString overloads (consistent with LTrim/RTrim)
- Behavior matches MySQL / standard SQL

## Verification

SELECT trim('  hello  ');   -- hello
SELECT lpad('hi', 5, 'x');  -- xxxhi
SELECT rpad('hi', 5, 'x');  -- hixxx

### DIFF
--- a/geaflow/geaflow-dsl/geaflow-dsl-plan/src/main/java/org/apache/geaflow/dsl/udf/table/string/LPad.java
+++ b/geaflow/geaflow-dsl/geaflow-dsl-plan/src/main/java/org/apache/geaflow/dsl/udf/table/string/LPad.java
@@ -1,0 +1,51 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.geaflow.dsl.udf.table.string;
+
+import org.apache.geaflow.dsl.common.function.Description;
+import org.apache.geaflow.dsl.common.function.UDF;
+
+@Description(name = "lpad", description = "Returns the string left-padded to the given length "
+    + "with the specified pad string.")
+public class LPad extends UDF {
+
+    public String eval(String str, Integer len, String pad) {
+        if (str == null || len == null || pad == null) {
+            return null;
+        }
+        if (len <= 0) {
+            return "";
+        }
+        if (len <= str.length()) {
+            return str.substring(0, len);
+        }
+        if (pad.isEmpty()) {
+            return str;
+        }
+        StringBuilder sb = new StringBuilder();
+        int padLen = len - str.length();
+        while (sb.length() < padLen) {
+            sb.append(pad);
+        }
+        sb.setLength(padLen);
+        sb.append(str);
+        return sb.toString();
+    }
+}
--- a/geaflow/geaflow-dsl/geaflow-dsl-plan/src/main/java/org/apache/geaflow/dsl/udf/table/string/RPad.java
+++ b/geaflow/geaflow-dsl/geaflow-dsl-plan/src/main/java/org/apache/geaflow/dsl/udf/table/string/RPad.java
@@ -1,0 +1,50 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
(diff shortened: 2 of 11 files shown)


candy972023:
#789

geaflow/geaflow-dsl/geaflow-dsl-plan/src/main/java/org/apache/geaflow/dsl/udf/table/string/LPad.java
```diff
@@ -0,0 +1,51 @@
...
+public class LPad extends UDF {
+
+    public String eval(String str, Integer len, String pad) {
+        if (str == null || len == null || pad == null) {
+            return null;
+        }
+        if (len <= 0) {
+            return "";
+        }
+        if (len <= str.length()) {
```
Leomrlin, line 36: `String.length()`, `substring()`, and `StringBuilder.length()` operate on UTF-16 code units rather than Unicode code points. This makes LPAD incorrect for supplementary characters and can split a surrogate pair while truncating.

  For example, `lpad("😀", 2, "x")` should return `"x😀"` under MySQL's character-based semantics, but this implementation returns `"😀"`.

geaflow/geaflow-dsl/geaflow-dsl-plan/src/main/java/org/apache/geaflow/dsl/udf/table/string/RPad.java
```diff
@@ -0,0 +1,50 @@
...
+public class RPad extends UDF {
+
+    public String eval(String str, Integer len, String pad) {
+        if (str == null || len == null || pad == null) {
+            return null;
+        }
+        if (len <= 0) {
+            return "";
+        }
+        if (len <= str.length()) {
```
Leomrlin, line 36: This has the same UTF-16 code-unit issue as LPAD.

geaflow/geaflow-dsl/geaflow-dsl-plan/src/main/java/org/apache/geaflow/dsl/udf/table/string/Trim.java
```diff
@@ -0,0 +1,51 @@
...
+import org.apache.geaflow.dsl.common.function.UDF;
+
+@Description(name = "trim", description = "Returns a string with both leading and trailing whitespace removed.")
+public class Trim extends UDF {
+
+    public String eval(String s) {
+        if (s == null) {
+            return null;
+        }
+        return StringUtils.strip(s);
```
Leomrlin, line 34: `StringUtils.strip(s)` removes tabs and other Java whitespace characters, while the `BinaryString` overload below removes only the ASCII space character. This creates overload-dependent behavior.

  MySQL's default TRIM behavior removes spaces, so `trim("\thello\t")` should retain the tabs, but this overload currently returns `"hello"`.

geaflow/geaflow-dsl/geaflow-dsl-runtime/src/test/resources/query/function_lpad.sql
```diff
@@ -0,0 +1,35 @@
...
+    c3 varchar,
+    c4 varchar
+) WITH (
+    type='file',
+    geaflow.dsl.file.path='${target}'
+);
+
+INSERT INTO output_console
+SELECT
+    lpad('hi', 5, 'x'),
```
Leomrlin, line 32: Could we extend the runtime tests with supplementary Unicode characters, null arguments, an empty padding string, and zero/negative target lengths?

geaflow/geaflow-dsl/geaflow-dsl-runtime/src/test/resources/query/function_trim.sql
```diff
@@ -0,0 +1,33 @@
...
+    c2 varchar,
+    c3 varchar
+) WITH (
+    type='file',
+    geaflow.dsl.file.path='${target}'
+);
+
+INSERT INTO output_console
+SELECT
+    trim('  hello  '),
```
Leomrlin, line 31: Please add a case containing tabs, such as `trim('\thello\t')`, to define and verify the intended semantics. It would also be useful to cover null input here.

Leomrlin (approved):

PR #816, opened 2026-07-22
feat(dsl): add MongoDB connector

### LINKED ISSUE #797: Add a MongoDB connector
### What & why

GeaFlow already supports JDBC / Hive / ES / Neo4j, but there's no connector for
the most popular document database, **MongoDB**. It's a frequent community
request — Mongo can serve as both a vertex/edge data source and a result sink.

GeaFlow 已支持 JDBC/Hive/ES/Neo4j,但缺主流文档数据库 **MongoDB**。这是社区高频需求:
Mongo 既能做点/边数据源,也能做结果输出。

### The task

Create a new module `geaflow-dsl-connector-mongodb` with a source (partitioned
reads, optional predicate pushdown) and a sink (batched writes).

新建模块 `geaflow-dsl-connector-mongodb`,实现 source(分区读、谓词下推可选)+ sink(批量写)。

### Where to look / 怎么做

1. Copy the most complete references for scaffolding: streaming
   `geaflow-dsl-connector-kafka`, partition/format `geaflow-dsl-connector-file`,
   remote-store class `geaflow-dsl-connector-elasticsearch`.
2. Implement `MongoTableConnector` (`TableReadableConnector` +
   `TableWritableConnector`), `MongoTableSource`
   (`init/open/listPartitions/getDeserializer/fetch`), and `MongoTableSink`
   (`init/open/write/flush/close`).
3. SPI registration:
   `src/main/resources/META-INF/services/org.apache.geaflow.dsl.connector.api.TableConnector`.
4. Add the module to the parent `pom.xml` and to the connector aggregation pom.
5. Docs (CN + EN) + an examples GQL + an integration test.

### Done when

- [ ] Both source and sink work, with basic connection config (uri / db / collection / batch size)
- [ ] `listPartitions` splits by shard/range to enable parallel reads
- [ ] Integration test included (Testcontainers can spin up MongoDB)
- [ ] CN + EN docs and one examples GQL
- [ ] Passes checkstyle and RAT (ASF header)

johntomcat7408-cmyk:

- Add a MongoDB source and sink connector with SPI registration.
- Support bounded reads, explicit integer range partitions with `[lowerbound, upperbound)`
  boundaries, and count offsets.
- Support ordered batch inserts and BSON type conversion.
- Add connector configuration, runtime wiring, unit and integration tests, a GQL example,
  and CN/EN docs.

The source targets static collections. CDC, upsert, automatic retry, and exactly-once delivery
are not included.

Closes #797.

### How was this PR tested?

- [x] Tests have been added for the changes

- `MongoTableConnectorTest` and `MongoRowConverterTest`: 15 tests passed on JDK 8 and JDK 11.
- `MongoTableConnectorIT`: passed with Docker in
  [GitHub Actions](https://github.com/johntomcat7408-cmyk/geaflow/actions/runs/29851217238);
  0 failures, errors, or skips.
- JDK 8 reactor build through the MongoDB connector passed: 55 modules.
- JDK 11 runtime reactor package passed: 75 modules, including Doris, MongoDB, and runtime.
- Checkstyle and Apache RAT passed.

### DIFF
--- a/geaflow/geaflow-dsl/geaflow-dsl-connector/geaflow-dsl-connector-mongodb/src/main/java/org/apache/geaflow/dsl/connector/mongodb/MongoTableSink.java
+++ b/geaflow/geaflow-dsl/geaflow-dsl-connector/geaflow-dsl-connector-mongodb/src/main/java/org/apache/geaflow/dsl/connector/mongodb/MongoTableSink.java
@@ -1,0 +1,123 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.geaflow.dsl.connector.mongodb;
+
+import com.mongodb.MongoException;
+import com.mongodb.client.MongoClient;
+import com.mongodb.client.MongoClients;
+import com.mongodb.client.MongoCollection;
+import com.mongodb.client.model.InsertManyOptions;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import org.apache.geaflow.api.context.RuntimeContext;
+import org.apache.geaflow.common.config.ConfigKey;
+import org.apache.geaflow.common.config.Configuration;
+import org.apache.geaflow.dsl.common.data.Row;
+import org.apache.geaflow.dsl.common.exception.GeaFlowDSLException;
+import org.apache.geaflow.dsl.common.types.StructType;
+import org.apache.geaflow.dsl.connector.api.TableSink;
+import org.bson.Document;
+
+public class MongoTableSink implements TableSink {
+
+    private String uri;
+    private String databaseName;
+    private String collectionName;
+    private int batchSize;
+    private MongoRowConverter converter;
+    private List<Document> batch;
+
+    private transient MongoClient client;
+    private transient MongoCollection<Document> collection;
+
+    @Override
+    public void init(Configuration tableConf, StructType schema) {
+        this.uri = required(tableConf, MongoConfigKeys.GEAFLOW_DSL_MONGODB_URI);
+        this.databaseName = required(tableConf, MongoConfigKeys.GEAFLOW_DSL_MONGODB_DATABASE);
+        this.collectionName = required(tableConf, MongoConfigKeys.GEAFLOW_DSL_MONGODB_COLLECTION);
+        this.batchSize = tableConf.getInteger(MongoConfigKeys.GEAFLOW_DSL_MONGODB_BATCH_SIZE);
+        if (batchSize <= 0) {
+            throw new GeaFlowDSLException("MongoDB batch size must be greater than zero");
+        }
+        this.converter = new MongoRowConverter(schema);
+        this.batch = new ArrayList<>(batchSize);
+    }
+
+    @Override
+    public void open(RuntimeContext context) {
+        try {
+            client = MongoClients.create(uri);
+            collection = client.getDatabase(databaseName).getCollection(collectionName);
+        } catch (MongoException | IllegalArgumentException e) {
+            close();
+            throw new GeaFlowDSLException("Failed to create MongoDB sink client", e);
+        }
+    }
+
+    @Override
+    public void write(Row row) throws IOException {
+        if (collection == null) {
+            throw new GeaFlowDSLException("MongoDB sink is not open");
+        }
+        batch.add(converter.toDocument(row));
+        if (batch.size() >= batchSize) {
+            flush();
+        }
+    }
+
+    @Override
+    public void finish() throws IOException {
+        flush();
+    }
+
+    @Override
+    public void close() {
+        if (client != null) {
+            client.close();
+            client = null;
+            collection = null;
+        }
+    }
+
+    private void flush() throws IOException {
+        if (batch.isEmpty()) {
+            return;
+        }
+        try {
+            collection.insertMany(batch, new InsertManyOptions().ordered(true));
+            batch.clear();
+        } catch (MongoException e) {
+            throw new IOException("Failed to write MongoDB collection " + collectionName, e);
+        }
+    }
+
+    private static String required(Configuration conf, ConfigKey key) {
+        if (!conf.contains(key)) {
+            throw new GeaFlowDSLException("Missing MongoDB configuration '{}'", key.getKey());
+        }
+        String value = conf.getString(key);
+        if (value == null || value.trim().isEmpty()) {
+            throw new GeaFlowDSLException("MongoDB configuration '{}' must not be blank",
+                key.getKey());
+        }
+        return value;
+    }
+}
--- a/geaflow/geaflow-dsl/geaflow-dsl-connector/geaflow-dsl-connector-mongodb/src/main/java/org/apache/geaflow/dsl/connector/mongodb/MongoTableSink.java
+++ b/geaflow/geaflow-dsl/geaflow-dsl-connector/geaflow-dsl-connector-mongodb/src/main/java/org/apache/geaflow/dsl/connector/mongodb/MongoTableSink.java
@@ -19,6 +19,7 @@
 
 package org.apache.geaflow.dsl.connector.mongodb;
 
+import com.mongodb.MongoBulkWriteException;
 import com.mongodb.MongoException;
 import com.mongodb.client.MongoClient;
 import com.mongodb.client.MongoClients;
@@ -104,6 +105,13 @@ public class MongoTableSink implements TableSink {
         try {
             collection.insertMany(batch, new InsertManyOptions().ordered(true));
             batch.clear();
+        } catch (MongoBulkWriteException e) {
+            String writeConcernError = e.getWriteConcernError() == null ? ""
+                : "; write concern error: " + e.getWriteConcernError();
+            throw new IOException("MongoDB bulk write failed for collection " + collectionName
+                + "; successful inserts in this batch: "
+                + e.getWriteResult().getInsertedCount() + "; write errors: "
+                + e.getWriteErrors() + writeConcernError, e);
         } catch (MongoException e) {
             throw new IOException("Failed to write MongoDB collection " + collectionName, e);
         }
--- a/geaflow/geaflow-dsl/geaflow-dsl-connector/geaflow-dsl-connector-mongodb/src/main/java/org/apache/geaflow/dsl/connector/mongodb/MongoTableSource.java
+++ b/geaflow/geaflow-dsl/geaflow-dsl-connector/geaflow-dsl-connector-mongodb/src/main/java/org/apache/geaflow/dsl/connector/mongodb/MongoTableSource.java
@@ -1,0 +1,220 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.geaflow.dsl.connector.mongodb;
+
+import com.mongodb.MongoException;
+import com.mongodb.client.FindIterable;
+import com.mongodb.client.MongoClient;
+import com.mongodb.client.MongoClients;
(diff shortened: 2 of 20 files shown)


johntomcat7408-cmyk:
I verified PR merge commit `b42bc097a2b144e0ae9713bb44af174d1cfb5983` (head `148d4791bf2f8199b0703077b26ecba7940d2713`) locally on Windows 11 with Oracle JDK 8u451, Maven 3.8.9, and protoc 3.21.7.

- All 107 modules compiled successfully, including test sources (`mvn ... -DskipTests`).
- MongoDB connector unit tests passed: 12 tests, 0 failures.

The failed JDK 8 job stopped at `setup-protoc` due to GitHub API rate limiting before Maven started. Could you please rerun it?

geaflow/geaflow-dsl/geaflow-dsl-connector/geaflow-dsl-connector-mongodb/src/main/java/org/apache/geaflow/dsl/connector/mongodb/MongoTableSource.java
```diff
@@ -0,0 +1,220 @@
...
+            }
+            limit = (int) windowSize;
+        }
+
+        List<Document> documents = new ArrayList<>();
+        try {
+            FindIterable<Document> query = collection.find(mongoPartition.toFilter())
+                .sort(mongoPartition.hasRange()
+                    ? Sorts.ascending(mongoPartition.getField(), "_id") : Sorts.ascending("_id"))
+                .skip((int) offset);
```
kitalkuyo-gita: Using `skip(offset)` for pagination poses potential performance risks.

### Issue
The `fetch()` method utilizes `query.skip((int) offset)`, relying on a count-based offset (where the offset represents the number of documents to skip). Combining `skip` with `sort` is highly inefficient when dealing with large collections or high offset values; this leads to full collection scans and memory pressure, with response times degrading linearly as the offset increases.

### Risks/Impact
With large datasets, this can result in extremely slow read operations, Out-of-Memory (OOM) errors, timeouts, or task failures. For dynamic collections, it may lead to duplicate records or missed data.

### Proposed Solution
Switch to range-based pagination (cursor-based) using the "last seen value." Instead of `skip`, use the sorting fields (e.g., `partitionField` + `_id`) to perform queries such as `> lastSeen`.
Extend the offset semantics to support two strategies: `COUNT_OFFSET` and `BOOKMARK_OFFSET`. While `COUNT_OFFSET` should be retained as the default for backward compatibility, it is strongly recommended to implement `BOOKMARK_OFFSET` (storing the `_id` or partition field value of the last returned record) to enable efficient pagination. ### Example:
New `MongoOffset` example (saving `lastId`):
```
public class MongoOffset implements Offset {
private final Long countOffset;          // Existing semantics (nullable)
private final String lastIdHex;          // New semantics (nullable)

public MongoOffset(long countOffset) { this.countOffset = countOffset; this.lastIdHex = null; }
public MongoOffset(String lastIdHex) { this.lastIdHex = lastIdHex; this.countOffset = null; }

public boolean isBookmark() { return lastIdHex != null; }
public String getLastIdHex() { return lastIdHex; }
@Override public long getOffset() { return countOffset == null ? 0L : countOffset; }
@Override public boolean isTimestamp() { return false; }
@Override public String humanReadable() { return lastIdHex != null ? lastIdHex : String.valueOf(countOffset); }
}
```
Prioritize bookmark-based pagination in `fetch()` (example snippet):

```
// When the offset is a bookmark (lastIdHex), use Filters.gt("_id", new ObjectId(lastIdHex))
Bson filter = mongoPartition.toFilter();
if (startOffset.isPresent() && startOffset.get() instanceof MongoOffset
&& ((MongoOffset) startOffset.get()).isBookmark()) {
String lastIdHex = ((MongoOffset) startOffset.get()).getLastIdHex();
filter = Filters.and(filter, Filters.gt("_id", new ObjectId(lastIdHex)));
} else {
// Fall back to the original skip(offset) behavior (for compatibility)
query = query.skip((int) offset);
}
// Apply limit + sort directly when no skip is used (efficient)
query = query.sort(Sorts.ascending(mongoPartition.hasRange() ? mongoPartition.getField() : "_id"))
.limit(limit);
```
When returning `nextOffset`, generate a `MongoOffset(lastIdHex)` using the `_id` of the last record; this allows the next `fetch` operation to use `> lastId` for pagination (O(1) overhead).

geaflow/geaflow-dsl/geaflow-dsl-connector/geaflow-dsl-connector-mongodb/src/main/java/org/apache/geaflow/dsl/connector/mongodb/MongoTableSink.java
```diff
@@ -0,0 +1,123 @@
...
+    public void close() {
+        if (client != null) {
+            client.close();
+            client = null;
+            collection = null;
+        }
+    }
+
+    private void flush() throws IOException {
+        if (batch.isEmpty()) {
```
kitalkuyo-gita, line 102: Using `ordered insertMany` in `flush()` results in partial writes without retries (impacting data consistency/availability).

### Issue
When using `ordered insertMany(true)`, an error causes the first portion of the batch to remain written while the method immediately throws an `IOException`. There is no mechanism for retries, rollbacks, or reporting granular details regarding which records were successfully written.

### Risks/Impact
Network instability or transient MongoDB backend issues can lead to partial writes and task failures, resulting in upstream duplication or downstream data loss (prone to user misinterpretation).

### Proposed Solution
Introduce configurable retries (max attempts, backoff strategy) and implement granular handling of `MongoBulkWriteException` upon insertion failure (optional: switch to unordered batch writes and retry failed sub-items based on error codes).
Provide configuration options: `geaflow.dsl.mongodb.sink.retry.maxAttempts`, `.retry.backoff.ms`, and `.sink.ordered` (true/false). Example implementation:
```
private void flushWithRetry() throws IOException {
if (batch.isEmpty()) return;
int attempts = 0;
int maxAttempts = 3; // Read from configuration
long backoffMs = 200L;
List<Document> toWrite = new ArrayList<>(batch);
while (true) {
try {
collection.insertMany(toWrite, new InsertManyOptions().ordered(true));
batch.clear();
return;
} catch (MongoBulkWriteException bwe) {
// If partial writes occurred, bwe.getWriteResult() is available, and bwe.getWriteErrors() returns the failed items.
// Log error info and retry failed items (but be mindful of idempotency and the risk of duplicate writes).
// Simple approach: Log and re-throw an exception with context (safer), or switch to unordered mode and retry failed items.
throw new IOException("Bulk write partially failed: " + bwe.getWriteErrors(), bwe);
} catch (MongoException e) {
attempts++;
if (attempts >= maxAttempts) {
throw new IOException("Failed to write MongoDB after retries", e);
}
try { Thread.sleep(backoffMs); } catch (InterruptedException ie) { Thread.currentThread().interrupt(); }
backoffMs *= 2;
}
}
}
```
Note: To implement reliable retries, you need to evaluate write idempotency (e.g., potential `_id` conflicts) and the consequences of unordered bulk writes (which may result in out-of-order execution).

geaflow/geaflow-dsl/geaflow-dsl-connector/geaflow-dsl-connector-mongodb/src/main/java/org/apache/geaflow/dsl/connector/mongodb/MongoTableSource.java
```diff
@@ -0,0 +1,220 @@
...
+
+    @Override
+    public List<Partition> listPartitions() {
+        if (partitionNum == 1) {
+            return Collections.singletonList(new MongoPartition(collectionName, 0));
+        }
+
+        BigInteger lower = BigInteger.valueOf(lowerBound);
+        BigInteger range = BigInteger.valueOf(upperBound).subtract(lower);
+        int count = range.min(BigInteger.valueOf(partitionNum)).intValueExact();
```
kitalkuyo-gita, line 114: `listPartitions` logic differs from expectations (`partitionNum` is effectively capped by the range size)

#### Issue
`listPartitions` uses `count = range.min(BigInteger.valueOf(partitionNum)).intValueExact();`. When the difference between the upper and lower bounds is less than `partitionNum`, the number of partitions actually returned is lower than the `partitionNum` specified by the user (this scenario is covered by tests in the PR). While this behavior is logical, it requires clear documentation and appropriate error or warning messages.

#### Risk/Impact
Users may specify a certain number of partitions to meet specific requirements; receiving fewer actual partitions for small ranges could lead to confusion regarding performance tuning.

#### Solution (Proposed)
The documentation should clarify this behavior, or a `WARN` log entry should be generated during initialization indicating that "requested partition count was reduced to X due to range size." Explicit checks or logging can be added to the code:

```
int requested = tableConf.getInteger(MongoConfigKeys.GEAFLOW_DSL_MONGODB_PARTITION_NUM);
if (count < requested) {
LOG.warn("Requested partitionNum={} reduced to {} because range size is {}.", requested, count, range);
}
```

geaflow/geaflow-dsl/geaflow-dsl-connector/geaflow-dsl-connector-mongodb/src/main/java/org/apache/geaflow/dsl/connector/mongodb/MongoTableSource.java
```diff
@@ -0,0 +1,220 @@
...
+                throw new GeaFlowDSLException("MongoDB fetch window size is out of range: {}",
+                    windowSize);
+            }
+            limit = (int) windowSize;
+        }
+
+        List<Document> documents = new ArrayList<>();
+        try {
+            FindIterable<Document> query = collection.find(mongoPartition.toFilter())
+                .sort(mongoPartition.hasRange()
```
kitalkuyo-gita, line 189: Partition/Pagination Sorting Requires Indexes (Performance)

#### Issue
`fetch()` utilizes `sort(mongoPartition.getField(), "_id")` or `sort("_id")`. Without a suitable compound index on the `partitionField` (or `_id`), the query results in costly sorting or scanning operations.

#### Risk/Impact
Queries on large collections may trigger disk-based sorting, out-of-memory errors, slow query performance, or even block the production cluster.

#### Solution (Recommendation)
Enforce and validate (during initialization or opening) that users create the necessary index on the collection (e.g., `{partitionField: 1, _id: 1}`)—or clearly document this requirement. Implement detection logic (using `.listIndexes()`) to log a warning or throw an exception, prompting the user to optimize the setup.

#### Example

```
// Check for the index after open(); issue a warning or optionally fail
boolean hasIndex = false;
for (Document idx : collection.listIndexes()) {
Document key = (Document) idx.get("key");
if (key.containsKey(partitionField) && key.containsKey("_id")) {
hasIndex = true;
break;
}
}
if (!hasIndex) {
LOG.warn("MongoDB collection {} missing index on ({}, _id). Recommended to create index for efficient reads.", collectionName, partitionField);
}
```

kitalkuyo-gita:
@johntomcat7408-cmyk  Hello! I’ve been extremely busy lately and have just seen your PR. Thank you very much for your valuable contribution to the community. We take a rigorous and serious approach with every new contributor; I have carefully reviewed your code and identified some performance issues, so I hope you can take a moment to look at them. Best regards.

geaflow/geaflow-dsl/geaflow-dsl-connector/geaflow-dsl-connector-mongodb/src/main/java/org/apache/geaflow/dsl/connector/mongodb/MongoTableSource.java
johntomcat7408-cmyk: Replaced count-based `skip` pagination with BSON bookmark offsets. Single-partition reads use `_id`; range-partitioned reads use `(partitionField, _id)` with the matching lexicographic filter and sort. Since this connector is new and unreleased, no count-offset fallback is retained. `ALL_WINDOW` reads now stream from the cursor.

geaflow/geaflow-dsl/geaflow-dsl-connector/geaflow-dsl-connector-mongodb/src/main/java/org/apache/geaflow/dsl/connector/mongodb/MongoTableSink.java
johntomcat7408-cmyk: Kept ordered, fail-fast writes. Connector-level retries or unordered writes are not added because a partially successful batch is not generally idempotent and retrying it can duplicate records. `MongoBulkWriteException` now reports successful inserts, write errors, and write concern errors. The delivery behavior is documented.

geaflow/geaflow-dsl/geaflow-dsl-connector/geaflow-dsl-connector-mongodb/src/main/java/org/apache/geaflow/dsl/connector/mongodb/MongoTableSource.java
johntomcat7408-cmyk: Documented `partition.num` as the maximum partition count. The actual count is `min(partition.num, upperbound - lowerbound)`, avoiding empty integer ranges.

geaflow/geaflow-dsl/geaflow-dsl-connector/geaflow-dsl-connector-mongodb/src/main/java/org/apache/geaflow/dsl/connector/mongodb/MongoTableSource.java
johntomcat7408-cmyk: Documented the recommended ascending compound index `{partitionField: 1, _id: 1}` for range-partitioned reads. Runtime index validation is not added because it would require extra permissions and a metadata query during source startup.

johntomcat7408-cmyk:
The follow-up changes are pushed in `f8a46870`. The [JDK 8 workflow](https://github.com/apache/geaflow/actions/runs/30429273791) and [JDK 11 workflow](https://github.com/apache/geaflow/actions/runs/30429273767) are waiting for approval. Could a maintainer approve them when convenient?

kitalkuyo-gita:
Great, thank you very much for the changes; CI has been started.

johntomcat7408-cmyk:
Both JDK 8 and JDK 11 CI runs have passed. Could you please take another look when convenient and let me know if any further changes are needed?

Leomrlin (approved):

kitalkuyo-gita (approved):
LGTM

Issue #801, opened 2026-06-18
Add the Betweenness Centrality graph algorithm


Leomrlin:
### What & why

GeaFlow has PageRank and ClosenessCentrality, but not **Betweenness Centrality** —
a measure of how often a vertex sits on shortest paths between other vertices,
i.e. how much of a "bridge" it is. It's a core metric for social-network and
key-node analysis. It's also genuinely hard: the classic Brandes algorithm has
two phases (shortest-path counting + dependency accumulation), which makes it a
satisfying high-difficulty challenge.

GeaFlow 有 PageRank、ClosenessCentrality,但缺**介数中心性**(Betweenness)——衡量一个
顶点作为"最短路径桥梁"的重要性,是社交网络/关键节点分析的核心指标。它实现难度大
(经典 Brandes 算法需要两阶段:最短路径计数 + 依赖累积),非常适合作为高难度挑战题。

### The task

Implement Betweenness Centrality on GeaFlow's vertex-centric / message-passing
model (an unweighted-graph version is a fine first cut).

基于 GeaFlow 的顶点中心/消息传递模型实现介数中心性(可先做无权图版本)。

### Where to look / 怎么做

1. Reference the more complex existing algorithms:
   `udf/graph/PageRank.java` (multi-iteration + convergence) and
   `SingleSourceShortestPath.java` (shortest paths).
2. Implement `AlgorithmUserFunction` using multi-round iteration + `sendMessage`
   for shortest-path counting and dependency back-propagation (the Brandes approach).
3. Register in `BuildInSqlFunctionTable.java`.
4. Use a small graph (5–7 nodes), compute the standard answer by hand, and turn
   it into `.sql`/`.txt` tests.

### Done when

- [ ] Implemented and registered; `CALL betweenness() YIELD (id, score)` works
- [ ] Results match hand-calculation / NetworkX on a small graph (attach the comparison in the PR)
- [ ] Complexity and suitable scale documented (note it's for small/medium graphs if so)
- [ ] Tests pass

PR #818, opened 2026-07-22
feat: support betweenness centrality algorithm (#801)

ArmxG:
What this PR does
Adds a built-in Betweenness Centrality graph algorithm (UDGA), as requested in #801. The algorithm computes the betweenness centrality score for every vertex and outputs `(id, betweenness)`.

Approach
This implementation uses a vertex-centric version of Brandes' algorithm with every vertex treated as a source.

* Forward phase: Performs BFS from each source to compute the number of shortest paths (`sigma`) and record predecessor information. ACK/NACK messages allow each vertex to determine when all child responses have been received.
* Backward phase: Propagates dependency values (`delta`) from the leaves back to the source, accumulating each vertex's betweenness centrality score.

Per-source state is stored in the vertex value, following the same design pattern as `AllSourceShortestPath`. The algorithm has O(V·E) time complexity and O(V²) space complexity.

Tests
* Added `gql_algorithm_betweenness.sql` and its expected output, verified against the standard Brandes algorithm.
* Added `testAlgorithmBetweennessCentrality` to `GQLAlgorithmTest`.


geaflow/geaflow-dsl/geaflow-dsl-plan/src/main/java/org/apache/geaflow/dsl/udf/graph/BetweennessCentrality.java
```diff
@@ -0,0 +1,270 @@
...
+        int outDegree = outEdges.size();
+
+        // Iteration 1: every vertex initializes itself as a source and scatters.
+        if (context.getCurrentIterationId() == 1L) {
+            double[] arr = new double[STATE_LEN];
+            arr[DIST] = 0;
+            arr[SIGMA] = 1;
+            arr[FORWARDED] = 1;
+            Map<String, Object> state = newState();
+            Map<Object, double[]> st = (Map<Object, double[]>) state.get("st");
```
kitalkuyo-gita, line 110: Thank you for your valuable contribution. I have some suggestions regarding this code and would like to discuss them with you.

Problem: The per-source states `st` (Map<Object, double[]>) and `preds` are kept in the vertex state. The current implementation does not release the corresponding entries after a source completes, causing them to accumulate and resulting in uncontrollable O(V^2) memory usage.

Suggestion: When a source's `BACKWARD_DONE` is set, immediately clear the entry for that source in `st` and `preds` (freeing up memory). It's possible to maintain a `touched` collection to only iterate over the sources actually involved (more efficient), but the minimum fix is ​​to remove it after completion.


Here is a code example for reference; you might find it helpful.
```
// --- Insert/Replace Front: Safely Read or Initialize State ---
Map<String, Object> state;

if (updatedValues ​​!= null && updatedValues.isPresent()) {

state = (Map<String, Object>) updatedValues.get().getField(0, null);

if (state == null) {

state = newState();

}
} else {

/ // The vertex doesn't have a state yet, safely initialize it (so subsequent logic can rely on a non-null state)

state = newState();

}
double bc = 0.0;

if (state.get("bc") != null) {

bc = (Double) state.get("bc");

} final Map<Object, double[]> st = (Map<Object, double[]>) state.get("st");
```

geaflow/geaflow-dsl/geaflow-dsl-plan/src/main/java/org/apache/geaflow/dsl/udf/graph/BetweennessCentrality.java
```diff
@@ -0,0 +1,270 @@
...
+        for (Object source : new ArrayList<>(st.keySet())) {
+            double[] arr = st.get(source);
+            if (arr[BACKWARD_DONE] == 1) {
+                continue;
+            }
+            boolean forwarded = arr[FORWARDED] == 1;
+            boolean childCountFinal = forwarded && arr[RESP_COUNT] >= outDegree;
+            if (childCountFinal && arr[RECV_CHILD] >= arr[CHILD_COUNT]) {
+                double sigma = arr[SIGMA];
+                double delta = sigma * arr[PENDING_SUM];
```
kitalkuyo-gita, line 200: Problem: Calculating `value = (1.0 + delta) / sigma` and `delta = sigma * pendingSum` lacks protection against `sigma == 0` (theoretically, `sigma == 0` represents unaccessed data or an abnormal count), leading to division by zero or NaN.

Recommendation: Perform a `sigma == 0` check (or `sigma < eps`) before using `sigma` for division; in this case, mark the source as complete and skip sending `backwardMsg`.

geaflow/geaflow-dsl/geaflow-dsl-plan/src/main/java/org/apache/geaflow/dsl/udf/graph/BetweennessCentrality.java
```diff
@@ -0,0 +1,270 @@
...
+        while (messages.hasNext()) {
+            List<Object> msg = messages.next();
+            int tag = ((Number) msg.get(0)).intValue();
+            Object source = msg.get(1);
+            if (tag == FWD) {
+                forwardBySource.computeIfAbsent(source, k -> new ArrayList<>()).add(msg);
+            } else if (tag == ACK) {
+                double[] arr = st.get(source);
+                if (arr != null) {
+                    arr[CHILD_COUNT]++;
```
kitalkuyo-gita, line 136: Robustness of ACK/NACK and Response Count

Issue: The logic for RESP_COUNT and CHILD_COUNT currently relies on `outDegree(outEdges.size())` as the expected number of responses. However, in scenarios with directed graphs, parallel edges, or different edge loading methods, this may not be a strict match, leading to premature or delayed entry into the backward phase.

Recommendation: Record "how many forward messages have been sent to neighbors (actual childCandidates)" instead of directly using `outDegree`, or determine whether all responses have been received based on the `touchedNeighbors` set.

kitalkuyo-gita:
Hello @ArmxG, thank you for your valuable contribution! I've been a bit busy lately and just saw your code. I'm leaving some comments for you to review. Please feel free to point out any issues you find unreasonable.

PR #819, opened 2026-07-23
[ISSUE-817] Add geaflow-dsl-connector-doris (Stream Load sink + partitioned source)

### LINKED ISSUE #817: Add a Doris connector (stream-load batch writes)
### What & why

Doris is a high-performance real-time OLAP database with sub-second queries,
high-concurrency real-time ingestion and upserts (Unique Key model, partial-column
updates), MySQL-protocol compatibility, and strong aggregation/JOIN. It is a natural
downstream analytics store for graph-computing results. You can connect today via the
JDBC connector (Doris speaks MySQL protocol), but it only does row-by-row INSERT and
misses Doris's high-throughput load path - Stream Load - so write throughput is poor.

Doris 是高性能实时 OLAP 数据库,具备亚秒级查询、高并发实时写入与更新(Unique Key 模型、
部分列更新)、MySQL 协议兼容以及出色的聚合/JOIN 能力,是图计算结果天然的下游分析存储。
现在可通过 JDBC connector 接入(Doris 兼容 MySQL 协议),但只能逐行 INSERT,未利用 Doris
的高吞吐导入通道 Stream Load,写吞吐差。

### The task

Create `geaflow-dsl-connector-doris`. Sink writes via Doris Stream Load (buffered, flush
on threshold); source supports parallel partitioned reads.

新建 `geaflow-dsl-connector-doris`。Sink 通过 Doris Stream Load 写入(攒批,达到阈值时
flush);source 支持按分区并行读。

### Where to look / 怎么做

1. Use `geaflow-dsl-connector-jdbc` as a baseline and spot the differences between
   "generic JDBC" and "Doris-specific".
2. Sink: buffer rows in `write()`, flush to Doris via Stream Load (HTTP PUT to FE) on
   threshold or `flush()`; use CSV/columnar payload for throughput.
3. SPI registration + parent/aggregation pom + docs + tests.

### Done when

- [ ] Sink throughput is clearly better than row-by-row JDBC (attach a simple benchmark in the PR)
- [ ] Source supports parallel partitioned reads
- [ ] Testcontainers integration test
- [ ] CN + EN docs + examples
- [ ] checkstyle / RAT pass

E2ern1ty:

Closes #817.

Add a new `geaflow-dsl-connector-doris` connector so graph-computing results can be written to / read from Apache Doris efficiently. Doris speaks the MySQL protocol so it is reachable today via the JDBC connector, but JDBC only does row-by-row INSERT and misses Doris's high-throughput load path — Stream Load. This connector adds it.

**Sink (Stream Load)**
- Buffers rows in `write()` and flushes a whole batch via a single HTTP PUT to the FE `/_stream_load` endpoint once the row-count or byte-size threshold is reached, or on window `finish()`.
- Follows the FE→BE 307 redirect (re-sending the body), does basic auth, retries, and validates the response `Status`.
- Default payload is **JSON** (Gson with `serializeNulls` + `disableHtmlEscaping`), which correctly handles newlines, quotes, backslashes, unicode, null and empty strings; CSV is available as a no-escape high-throughput option.
- Multiple FEs can be configured; a failed request fails over to the next FE.

**Source (partitioned reads)**
- Reads through the MySQL protocol and splits the scan into N partitions on a numeric column for parallel reads; the first partition also covers rows whose partition column is entirely NULL.
- The GPL-licensed MySQL driver is **not bundled** (test scope only, like `geaflow-dsl-connector-jdbc`); users provide a MySQL-protocol driver on the runtime classpath.

Also: SPI registration, parent/aggregation pom + runtime wiring, and CN + EN docs with options, examples and benchmark numbers.

Addresses the maintainer's review points: safe CSV/JSON serialization; idempotent `finish()` and empty-buffer-safe flush; `close()` flushes the remaining buffer and is safe on repeated calls; `partition.num` is clamped; globally-unique labels via UUID; FE failover; and integration tests covering FE failover, table-not-found and full data-type round-trip fidelity.

### How was this PR tested?

- Unit tests (run in CI): serialization of special chars / null / empty string, config defaults, multi-FE url building, sink init validation. `checkstyle` 0 violations, `RAT` 0 unapproved.
- Testcontainers integration test (opt-in, disabled by default so CI stays green): sink+source round-trip, FE failover, table-not-found, full data-type fidelity, and a throughput benchmark.
- Verified end to end against a real single-node `apache/doris:3.0.8-all`: all four core integration cases pass. Benchmark: Stream Load loaded 500 rows in ~91ms (~5500 rows/s) vs ~11335ms (~44 rows/s) for row-by-row JDBC INSERT (~125x faster); 2000 rows via Stream Load took only ~97ms.

### DIFF
--- a/geaflow/geaflow-dsl/geaflow-dsl-connector/geaflow-dsl-connector-doris/src/main/java/org/apache/geaflow/dsl/connector/doris/DorisStreamLoad.java
+++ b/geaflow/geaflow-dsl/geaflow-dsl-connector/geaflow-dsl-connector-doris/src/main/java/org/apache/geaflow/dsl/connector/doris/DorisStreamLoad.java
@@ -1,0 +1,181 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.geaflow.dsl.connector.doris;
+
+import com.google.gson.Gson;
+import com.google.gson.JsonObject;
+import java.io.Closeable;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.util.Base64;
+import java.util.List;
+import java.util.UUID;
+import org.apache.geaflow.dsl.common.exception.GeaFlowDSLException;
+import org.apache.http.HttpHeaders;
+import org.apache.http.client.config.RequestConfig;
+import org.apache.http.client.methods.CloseableHttpResponse;
+import org.apache.http.client.methods.HttpPut;
+import org.apache.http.entity.ByteArrayEntity;
+import org.apache.http.impl.client.CloseableHttpClient;
+import org.apache.http.impl.client.DefaultRedirectStrategy;
+import org.apache.http.impl.client.HttpClients;
+import org.apache.http.util.EntityUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A lightweight Doris Stream Load client. It issues an HTTP PUT to the Doris FE, which replies
+ * with a 307 redirect to a BE, and the request body is re-sent to the BE that finally executes
+ * the load. The buffered payload is loaded in a single request, giving a much higher throughput
+ * than row-by-row JDBC INSERT.
+ */
+public class DorisStreamLoad implements Closeable {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(DorisStreamLoad.class);
+
+    private final String loadUrl;
+    private final String authHeader;
+    private final String format;
+    private final String columnSeparator;
+    private final String lineDelimiter;
+    private final String columns;
+    private final int maxRetries;
+    private final CloseableHttpClient httpClient;
+    private final Gson gson = new Gson();
+
+    public DorisStreamLoad(String feNode, String database, String table, String username,
+                           String password, String format, String columnSeparator,
+                           String lineDelimiter, List<String> columns, int connectTimeoutMs,
+                           int readTimeoutMs, int maxRetries) {
+        this.loadUrl = String.format(DorisConstants.STREAM_LOAD_URL_PATTERN,
+            normalizeFeNode(feNode), database, table);
+        this.authHeader = "Basic " + Base64.getEncoder().encodeToString(
+            (username + ":" + password).getBytes(StandardCharsets.UTF_8));
+        this.format = format;
+        this.columnSeparator = columnSeparator;
+        this.lineDelimiter = lineDelimiter;
+        this.columns = String.join(DorisConstants.COMMA, columns);
+        this.maxRetries = Math.max(1, maxRetries);
+        RequestConfig requestConfig = RequestConfig.custom()
+            .setConnectTimeout(connectTimeoutMs)
+            .setSocketTimeout(readTimeoutMs)
+            .setConnectionRequestTimeout(connectTimeoutMs)
+            .build();
+        // The FE returns a 307 redirect that must be followed with the same method and body,
+        // so mark every method as redirectable.
+        this.httpClient = HttpClients.custom()
+            .setDefaultRequestConfig(requestConfig)
+            .setRedirectStrategy(new DefaultRedirectStrategy() {
+                @Override
+                protected boolean isRedirectable(String method) {
+                    return true;
+                }
+            })
+            .build();
+    }
+
+    private static String normalizeFeNode(String feNode) {
+        String node = feNode.trim();
+        if (!node.startsWith(DorisConstants.HTTP_SCHEME) && !node.startsWith("https://")) {
+            node = DorisConstants.HTTP_SCHEME + node;
+        }
+        return node;
+    }
+
+    /**
+     * Load the given payload into Doris via a single Stream Load request. The method retries on
+     * transient failures and throws a {@link GeaFlowDSLException} once all retries are exhausted.
+     */
+    public void load(byte[] payload) {
+        Exception lastError = null;
+        for (int attempt = 1; attempt <= maxRetries; attempt++) {
+            try {
+                doLoad(payload);
+                return;
+            } catch (Exception e) {
+                lastError = e;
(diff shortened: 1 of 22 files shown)


geaflow/geaflow-dsl/geaflow-dsl-connector/geaflow-dsl-connector-doris/src/main/java/org/apache/geaflow/dsl/connector/doris/DorisTableSource.java
```diff
@@ -0,0 +1,300 @@
...
+            throw new GeaFlowDSLException("failed to load mysql driver for doris source.", e);
+        }
+    }
+
+    private String qualifiedTable() {
+        return database == null || database.isEmpty() ? table : database + "." + table;
+    }
+
+    @Override
+    public List<Partition> listPartitions() {
```
kitalkuyo-gita, line 126: Problem/Cause: Currently, only numeric range partitioning is supported. When columns are highly skewed, have no indexes, or are non-numeric (string/date), the partitioning strategy may lead to hotspots or uneven partitioning.

Recommendation: Provide multiple partitioning strategies (range, hash, sampling-based), allow users to manually specify partition boundaries, or use custom SQL subqueries. Emphasize the performance impact of unindexed partitioning columns in the documentation and recommend indexing/appropriate partitioning columns.

geaflow/geaflow-dsl/geaflow-dsl-connector/geaflow-dsl-connector-doris/src/main/java/org/apache/geaflow/dsl/connector/doris/DorisStreamLoad.java
```diff
@@ -0,0 +1,199 @@
...
+        String status = result.get(DorisConstants.STREAM_LOAD_RESULT_STATUS).getAsString();
+        if (!DorisConstants.STREAM_LOAD_SUCCESS.equals(status)
+            && !DorisConstants.STREAM_LOAD_PUBLISH_TIMEOUT.equals(status)) {
+            String message = result.has(DorisConstants.STREAM_LOAD_RESULT_MESSAGE)
+                ? result.get(DorisConstants.STREAM_LOAD_RESULT_MESSAGE).getAsString() : body;
+            throw new IOException("Stream Load failed, status: " + status + ", message: " + message);
+        }
+    }
+
+    private String generateLabel() {
```
kitalkuyo-gita, line 193: Partial Success and Duplicate Writes (Label/Idempotency)

Problem/Cause: Stream Load uses labels/UUIDs for assistance, but after FE failover and retries, there is a risk of duplicate writes or semantic inconsistencies due to "partial success + retries" (Doris's behavior for the same label needs to be explicitly defined).

Recommendation: Clearly define the semantics of labels in the README and implementation (if the server supports idempotency/duplicate checks), and:
Log the label in the failure/retry strategy and check the status of the label on the server before retrying (if it is queryable).

Provide optional "idempotency/duplicate removal strategies" (e.g., allow-duplicates vs. check-label), or expose the number of retries and their behavior as configuration options.

geaflow/geaflow-dsl/geaflow-dsl-connector/geaflow-dsl-connector-doris/src/main/java/org/apache/geaflow/dsl/connector/doris/DorisStreamLoad.java
```diff
@@ -0,0 +1,199 @@
...
+     * Load the given payload into Doris via a single Stream Load request. The method retries on
+     * transient failures and throws a {@link GeaFlowDSLException} once all retries are exhausted.
+     */
+    public void load(byte[] payload) {
+        Exception lastError = null;
+        for (int attempt = 0; attempt < maxRetries; attempt++) {
+            // Rotate over the FE list so a failed request fails over to the next FE.
+            String url = loadUrls.get(attempt % loadUrls.size());
+            try {
+                doLoad(url, payload);
```
kitalkuyo-gita: Connection/Read/Write Timeout and Large Payload Timeout

Problem/Cause: Stream Load request bodies can be very large. A default timeout that is too short can cause interruption; a timeout that is too long can easily exhaust resources in the event of network anomalies.

Recommendation: Set reasonable connection and read/write timeouts, and allow users to configure them. Provide chunked/streaming uploads for large requests (if Doris supports this) or add upload timeout instructions.

kitalkuyo-gita:
Hello @E2ern1ty, I'm glad you contributed to the community! I left some comments, please take a look and see if they need any changes.

geaflow/geaflow-dsl/geaflow-dsl-connector/geaflow-dsl-connector-doris/src/main/java/org/apache/geaflow/dsl/connector/doris/DorisTableSource.java
E2ern1ty: Thanks. Added a `source.partition.mode` option (`range`/`custom`). `custom` takes semicolon-separated WHERE predicates via `source.partition.clauses` (one partition per predicate), which supports non-numeric/skewed columns and arbitrary conditions (e.g. date subranges), so users can specify partition boundaries manually. Kept numeric `range` as the default. Docs now warn that the range partition column should be indexed and evenly distributed (otherwise full scans/hotspots) and recommend the primary key. Added DorisTableSourceTest covering single/range/custom listing. Sampling-based auto splitting can be a follow-up.

geaflow/geaflow-dsl/geaflow-dsl-connector/geaflow-dsl-connector-doris/src/main/java/org/apache/geaflow/dsl/connector/doris/DorisStreamLoad.java
E2ern1ty: Fixed the idempotency risk: the label is now generated once per batch and reused across every retry and FE failover attempt (previously a new label was created per attempt, which defeated Doris's dedup). Since Doris loads a given label at most once, a retry after a mistaken failure now hits "Label Already Exists", which is treated as success instead of writing the batch twice. The label is logged on every failure/retry. Documented these write semantics in the CN/EN docs. Retry behavior is configurable via `sink.max.retries` (kept at least one attempt per FE for failover).

geaflow/geaflow-dsl/geaflow-dsl-connector/geaflow-dsl-connector-doris/src/main/java/org/apache/geaflow/dsl/connector/doris/DorisStreamLoad.java
E2ern1ty: Connect and read/write timeouts are both user-configurable (`request.connect.timeout.ms`, `request.read.timeout.ms`) and applied to the HttpClient RequestConfig. Raised the default read/write timeout to 60s so it covers uploading a large batch, and documented tuning guidance (increase for very large batches on slow networks, lower to fail faster on anomalies). The payload is already sent as a single streamed HTTP entity; batch size itself is bounded by `sink.batch.rows`/`sink.batch.bytes` so a batch stays within a reasonable size.

kitalkuyo-gita (approved):
LGTM

Leomrlin (approved):
LGTM

PR #823, opened 2026-08-02
[ISSUE-796] Clean up PartitionType TODOs in rocksdb store module

### LINKED ISSUE #796: Clean up a small TODO in the store module
### What & why

The `geaflow-store` module has several small, well-scoped TODOs that make a great
doorway into the storage layer. For example,
`geaflow-store-rocksdb/.../PartitionType.java` has
`// TODO: Support dt partition` and `// TODO: Support label dt partition`, and
the API layer's `OneDegreeGraphScanIterator` has one too. Small scope, clear context.

`geaflow-store` 模块里有若干小 TODO,是读懂存储层的好切入点。比如
`PartitionType.java` 里的 `// TODO: Support dt partition`,以及 api 层
`OneDegreeGraphScanIterator` 的 TODO。范围小、上下文清晰。

### The task

Claim one TODO and either implement it, or — if it's genuinely not ready to
implement — replace it with a clear validation + error message plus a doc note.
Prefer the ones that are well-scoped and testable.

认领其中一个 TODO,补全实现;或(若暂不宜实现)替换成清晰的校验与报错 + 文档说明。
优先选范围明确、能加测试验证的。

### Where to look

1. Find the TODOs: `grep -rn "TODO" geaflow/geaflow-plugins/geaflow-store/`.
2. Mirror a similar, already-implemented branch in the same module.
3. Add a unit test covering the new path.

### Done when

- [ ] The chosen TODO is implemented, or replaced by an explicit "not supported"
      check (throwing a clear exception rather than failing silently)
- [ ] A test covers the new behavior
- [ ] Existing store tests still pass (`mvn test -pl <store-module>`)
- [ ] The PR describes the scope of the change

yyyCode:

Closes #796.

Cleans up two stale `TODO`s on the `PartitionType` enum in the rocksdb graph store (`geaflow-store-rocksdb`):

- **`DT` — `// TODO: Support dt partition`**: DT partition is already implemented via `SyncGraphDtPartitionProxy` and dispatched by `ProxyBuilder`, so this TODO is obsolete and is removed.
- **`DT_LABEL` — `// TODO: Support label dt partition`**: this combination has no proxy implementation. Previously, configuring `partition.type=dt_label` fell through to a generic `"unexpected partition type"` error. `ProxyBuilder.build` now rejects `DT_LABEL` explicitly with a message that names the unsupported type (fail-fast), and the TODO is replaced with a `NOTE` documenting the gap for future implementers.

No behavior change for the supported partition types (`NONE`, `LABEL`, `DT`).

### How was this PR tested?

Added `PartitionTypeTest` covering:
- case-insensitive enum lookup and rejection of unknown types;
- the partition flags of `DT` / `DT_LABEL` / `NONE`;
- `ProxyBuilder` rejecting `DT_LABEL` with an explicit message that names the type;
- `ProxyBuilder` still dispatching `DT` to a real proxy.

The test targets `ProxyBuilder` directly, so it does not require opening a native RocksDB instance. `mvn test -Dtest=PartitionTypeTest` passes (5/5) and `mvn checkstyle:check` reports 0 violations on the module.

### DIFF
--- a/geaflow/geaflow-plugins/geaflow-store/geaflow-store-rocksdb/src/main/java/org/apache/geaflow/store/rocksdb/proxy/ProxyBuilder.java
+++ b/geaflow/geaflow-plugins/geaflow-store/geaflow-store-rocksdb/src/main/java/org/apache/geaflow/store/rocksdb/proxy/ProxyBuilder.java
@@ -40,6 +40,10 @@ public class ProxyBuilder {
                 return new SyncGraphLabelPartitionProxy<>(rocksdbClient, encoder, config);
             } else if (partitionType == PartitionType.DT) {
                 return new SyncGraphDtPartitionProxy<>(rocksdbClient, encoder, config);
+            } else if (partitionType == PartitionType.DT_LABEL) {
+                throw new GeaflowRuntimeException(
+                    "partition type DT_LABEL (partition by both timestamp and label) is not "
+                        + "supported yet");
             }
             throw new GeaflowRuntimeException("unexpected partition type: " + config.getString(
                 RocksdbConfigKeys.ROCKSDB_GRAPH_STORE_PARTITION_TYPE));
--- a/geaflow/geaflow-plugins/geaflow-store/geaflow-store-rocksdb/src/main/java/org/apache/geaflow/store/rocksdb/proxy/ProxyBuilder.java
+++ b/geaflow/geaflow-plugins/geaflow-store/geaflow-store-rocksdb/src/main/java/org/apache/geaflow/store/rocksdb/proxy/ProxyBuilder.java
@@ -29,21 +29,35 @@ import org.apache.geaflow.store.rocksdb.RocksdbConfigKeys;
 
 public class ProxyBuilder {
 
+    /**
+     * Rejects partition types that have no proxy implementation yet. Callers should invoke this
+     * before opening the RocksDB instance so an unsupported {@code partition.type} fails fast
+     * without leaving a half-initialized store on disk.
+     *
+     * <p>{@code DT_LABEL} (partition by both timestamp and label) has no proxy, so it is rejected
+     * with an explicit message that names the type rather than falling through to a generic
+     * "unexpected partition type" error later in {@link #build}.
+     */
+    public static void checkPartitionTypeSupported(PartitionType partitionType) {
+        if (partitionType == PartitionType.DT_LABEL) {
+            throw new GeaflowRuntimeException(
+                "partition type DT_LABEL (partition by both timestamp and label) is not "
+                    + "supported yet");
+        }
+    }
+
     public static <K, VV, EV> IGraphRocksdbProxy<K, VV, EV> build(
         Configuration config, RocksdbClient rocksdbClient,
         IGraphKVEncoder<K, VV, EV> encoder) {
         PartitionType partitionType = PartitionType.getEnum(
             config.getString(RocksdbConfigKeys.ROCKSDB_GRAPH_STORE_PARTITION_TYPE));
+        checkPartitionTypeSupported(partitionType);
         if (partitionType.isPartition()) {
             if (partitionType == PartitionType.LABEL) {
                 // TODO: Support async graph proxy partitioned by label
                 return new SyncGraphLabelPartitionProxy<>(rocksdbClient, encoder, config);
             } else if (partitionType == PartitionType.DT) {
                 return new SyncGraphDtPartitionProxy<>(rocksdbClient, encoder, config);
-            } else if (partitionType == PartitionType.DT_LABEL) {
-                throw new GeaflowRuntimeException(
-                    "partition type DT_LABEL (partition by both timestamp and label) is not "
-                        + "supported yet");
             }
             throw new GeaflowRuntimeException("unexpected partition type: " + config.getString(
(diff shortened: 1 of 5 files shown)


geaflow/geaflow-plugins/geaflow-store/geaflow-store-rocksdb/src/main/java/org/apache/geaflow/store/rocksdb/proxy/ProxyBuilder.java
```diff
@@ -40,6 +40,10 @@ public static <K, VV, EV> IGraphRocksdbProxy<K, VV, EV> build(
                 return new SyncGraphLabelPartitionProxy<>(rocksdbClient, encoder, config);
             } else if (partitionType == PartitionType.DT) {
                 return new SyncGraphDtPartitionProxy<>(rocksdbClient, encoder, config);
+            } else if (partitionType == PartitionType.DT_LABEL) {
```
Leomrlin: DT_LABEL is rejected here, but this is not fail-fast in the actual store initialization path

geaflow/geaflow-plugins/geaflow-store/geaflow-store-rocksdb/src/test/java/org/apache/geaflow/store/rocksdb/PartitionTypeTest.java
```diff
@@ -0,0 +1,91 @@
...
+        }
+    }
+
+    /**
+     * DT partition is implemented (see {@code SyncGraphDtPartitionProxy}), so {@link ProxyBuilder}
+     * must dispatch it to a real proxy instead of rejecting it. With a {@code null} client the
+     * proxy constructor fails with a {@link NullPointerException}, which confirms the builder got
+     * past partition-type dispatch rather than throwing a {@link GeaflowRuntimeException}.
+     */
+    @Test(expectedExceptions = NullPointerException.class)
```
Leomrlin: NullPointerException does not clearly indicate the issue

geaflow/geaflow-plugins/geaflow-store/geaflow-store-rocksdb/src/main/java/org/apache/geaflow/store/rocksdb/proxy/ProxyBuilder.java
yyyCode: Good catch. The real path is StaticGraphRocksdbStoreBase.init -> super.init() (which opens the native RocksDB in BaseRocksdbStore.init) -> ProxyBuilder.build(), so rejecting DT_LABEL only inside build() means the DB is already opened on disk before we fail.

Fixed in 5b01c37d: extracted ProxyBuilder.checkPartitionTypeSupported(PartitionType) and call it in StaticGraphRocksdbStoreBase.init *before* super.init(), so an unsupported partition.type now fails fast without leaving a half-initialized store on disk. build() still calls the same check as well.

geaflow/geaflow-plugins/geaflow-store/geaflow-store-rocksdb/src/test/java/org/apache/geaflow/store/rocksdb/PartitionTypeTest.java
yyyCode: Agreed, inferring dispatch from an incidental NullPointerException is fragile and does not state the intent. Fixed in 5b01c37d: the test now mocks RocksdbClient/encoder (mockito-all, version managed by the parent pom) and asserts the returned proxy is a SyncGraphDtPartitionProxy, so it directly verifies that DT is dispatched to the DT proxy.

geaflow/geaflow-plugins/geaflow-store/geaflow-store-rocksdb/src/main/java/org/apache/geaflow/store/rocksdb/proxy/ProxyBuilder.java
yyyCode: Thanks for the review @Leomrlin. Verified locally on the latest commit: an unsupported `partition.type` now fails in `StaticGraphRocksdbStoreBase.init` *before* `super.init()` opens the native RocksDB, so no half-initialized store is left on disk, and `ProxyBuilder.build()` keeps the same guard. `mvn -Dtest=PartitionTypeTest test` passes 5/5. Could you take another look when you have a moment?

geaflow/geaflow-plugins/geaflow-store/geaflow-store-rocksdb/src/test/java/org/apache/geaflow/store/rocksdb/PartitionTypeTest.java
yyyCode: Thanks @Leomrlin. Updated so the test asserts `proxy instanceof SyncGraphDtPartitionProxy` directly (mocking `RocksdbClient`/encoder) instead of inferring dispatch from an incidental `NullPointerException`, which makes the intent explicit. Verified locally: `PartitionTypeTest` passes 5/5. PTAL when you get a chance.

Leomrlin (approved):
LGTM

kitalkuyo-gita (approved):
LGTM

PR #675, opened 2025-11-17
feat(dsl): 实现ISO-GQL源/目标谓词函数

SeasonPilot:
https://github.com/apache/geaflow/issues/366

### What changes were proposed in this pull request?

- 新增SourceDestinationFunctions工具类，实现ISO-GQL标准的源/目标谓词逻辑
- 添加四个新的UDF函数：IsSourceOf、IsNotSourceOf、IsDestinationOf、IsNotDestinationOf
- 在BuildInSqlFunctionTable中注册新的源/目标谓词函数
- 实现完整的三值逻辑处理，包括空值和无向边情况
- 添加五个测试用例验证源/目标谓词的正确性
- 提供详细的JavaDoc说明和ISO-GQL规则实现注释
### How was this PR tested?
- [x] Tests have Added for the changes
- [x] Production environment verified

PR #824, opened 2026-08-02
[ISSUE-358] Implement ISO-GQL labeled predicate

### LINKED ISSUE #358: Added implementation of standard ISO-GQL syntax(1)：labeled predicate
【需求描述】
新增实现以下标准ISO-GQL语法:
1. labeled predicate

【期望产出】
1. 支持实现上述语法，同时提供并跑通测试用例。 

如有疑问，请联系：@puuuuug

yyyCode:

Closes #358.

Implements the ISO-GQL `<labeled predicate>` (ISO/IEC 39075 Section 19.9), which tests whether a graph element (vertex or edge) carries a given label.

Following the pattern of the already-merged **source/destination predicate** (#366 / PR #675), this is implemented as built-in UDFs rather than a grammar change:

- **`IS_LABELED(element, label)`** — returns TRUE if the vertex/edge has the given label, FALSE otherwise.
- **`IS_NOT_LABELED(element, label)`** — the negation.

Both follow ISO-GQL **three-valued logic**: a null element or null label yields Unknown (`null`). A non-graph-element first operand raises a clear `IllegalArgumentException`.

Files:
- `LabeledPredicateFunctions` (geaflow-dsl-common) — shared static logic.
- `IsLabeled` / `IsNotLabeled` (geaflow-dsl-plan) — UDFs delegating to the logic, registered in `BuildInSqlFunctionTable`.

**Example:**
```sql
MATCH (a) -[e]-> (b)
WHERE IS_LABELED(a, 'person')
RETURN a, e, b
```

### How was this PR tested?

- **`LabeledPredicateFunctionsTest`** (geaflow-dsl-common): unit-tests the logic directly — vertex/edge label matching, negation, three-valued null handling, and rejection of non-graph-element operands. Passes 5/5 locally.
- **`GQLLabeledPredicateTest`** (geaflow-dsl-runtime): an end-to-end `.sql`/`.txt` case over the `modern` graph, exercising `IS_LABELED`/`IS_NOT_LABELED` on both vertices and edges. Expected outputs were computed by hand from the graph data (not copied from program output).

`mvn checkstyle:check` reports 0 violations across the three touched modules.

> Note: the runtime end-to-end test could not be executed on my local Windows machine because Hadoop's `winutils.exe` is required to write the file sink (the existing `GQLSourceDestinationTest` fails locally for the same reason). It is expected to run on the Linux CI. The predicate logic itself is fully covered by the platform-independent unit test.


geaflow/geaflow-dsl/geaflow-dsl-common/src/main/java/org/apache/geaflow/dsl/common/function/LabeledPredicateFunctions.java
yaozhongq, line 1: This package is intended for general-purpose interfaces. There is no need to split the current function into a separate `Func`; I suggest placing it directly into the UDF.

yyyCode: Good point, thanks. Fixed in 5c65fc85: I moved the label-matching and three-valued-null logic (plus a private getLabel helper) directly into IsLabeled, and IsNotLabeled now delegates to an IsLabeled instance and negates the result, so label extraction stays in one place. Removed LabeledPredicateFunctions and its test from geaflow-dsl-common, and moved the unit test into geaflow-dsl-plan as LabeledPredicateTest (same udf/table/other package as the UDFs, matching PropertyExistsTest), testing through the UDF eval methods.

Note: the existing merged predicates (IsSourceOf/PropertyExists) still use the split helper pattern this PR was originally modeled on, so the labeled predicate is now self-contained and diverges from them. Happy to align those too in a follow-up if you would prefer consistency.

yyyCode: Thanks for the review @yaozhongq. Confirmed on the latest commit: `LabeledPredicateFunctions` is gone from geaflow-dsl-common, the label-matching + three-valued-null logic (and the `getLabel` helper) now lives directly in `IsLabeled`, and `IsNotLabeled` delegates to an `IsLabeled` instance so label extraction stays in one place. The unit test moved to geaflow-dsl-plan as `LabeledPredicateTest` (same package as the UDFs, matching `PropertyExistsTest`) and passes 5/5 locally. PTAL when you get a chance.

yaozhongq: LGTM, looks great!

yaozhongq (approved):
LGTM

PR #829, opened 2026-08-04
test(dsl): add modern-graph regression tests for built-in graph algorithms

### LINKED ISSUE #794: Add runnable `.sql` examples and regression tests for the built-in graph algorithms
### What & why

GeaFlow ships 19 built-in graph algorithms (PageRank, SSSP, WCC, Louvain, LPA,
KCore, TriangleCount, JaccardSimilarity, …), but the `geaflow-dsl-runtime`
regression suite only covers some of them. Filling in end-to-end
`CALL ... YIELD` tests for the rest guards against future refactors quietly
breaking an algorithm. Great if you want to *understand* what graph algorithms
actually compute.

GeaFlow 有 19 个内置图算法,但回归测试只覆盖了一部分。给其余算法补端到端的
`CALL ... YIELD` 测试,能防止后续重构悄悄改坏算法。想真正搞懂"图算法到底算了什么"的同学,这题很合适。

### The task

Pick a few algorithms that have no tests (or thin coverage) and add `.sql`
queries + `.txt` expected outputs. 

挑几个尚无测试或覆盖薄弱的算法,补 `.sql` 查询 + `.txt` 期望输出。

### Where to look

1. Test graph definition: `geaflow-dsl-runtime/src/test/resources/modern_graph.sql`.
2. Query template: existing `query/gql_algorithm_001.sql`
   (`CALL SSSP(1) YIELD (vid, distance)` style); expected output at `expect/gql_algorithm_001.txt`.
3. Test entry point: `geaflow-dsl-runtime/.../runtime/query/GQLAlgorithmTest.java`.

### Done when

- [ ] At least one `.sql`/`.txt` pair added per algorithm you claim
- [ ] Expected outputs are **verified by hand**, not copied from program output
- [ ] `mvn test -Dtest=GQLAlgorithmTest` passes
- [ ] The PR briefly explains why each expected result is correct on the test graph

CalebWang0126:
Fixes #794

## What & why

Several built-in algorithms were never exercised on the standard `modern` graph by `GQLAlgorithmTest`: `khop`, `closeness_centrality` and `lpa` were only tested on ad-hoc graphs (g4/g5), and `common_neighbors` / `jaccard_similarity` had only a single parameter case. This PR adds five end-to-end `CALL ... YIELD` pairs on `modern_graph.sql`.

## Expected outputs — verified by hand

The modern graph has vertices 1..6 and directed edges 1->2, 1->3, 1->4, 4->3, 4->5, 6->3 (algorithms using `EdgeDirection.BOTH` treat it as undirected).

- **khop(1, 2)** -> `(1,0),(2,1),(3,1),(4,1),(5,2)`: BFS distances from vertex 1 within 2 hops; vertex 6 is unreachable. Consistent with the existing SSSP expectations on the same graph.
- **closeness_centrality(1)** -> `1.0`: sum of shortest distances from 1 to reachable vertices = 1+1+1+2 = 5, and n-1 = 5, so 5/5 = 1.0.
- **lpa()** -> all vertices converge to label `1`: simulated round by round; on ties the implementation keeps the lexicographically smallest label, which makes the outcome deterministic on this graph.
- **common_neighbors(4, 6)** -> `3`: N(4) = {1,3,5}, N(6) = {3}, intersection = {3}.
- **jaccard_similarity(1, 4)** -> `0.2`: N(1) = {2,3,4}, N(4) = {1,3,5}, intersection = {3}, union size = 5, so 1/5 = 0.2.

## Test

`mvn test -Dtest=GQLAlgorithmTest` passes (39 tests, including the 5 new ones). Checkstyle passes.

## Notes

While adding the jaccard case I noticed that for **non-adjacent** vertex pairs the algorithm always returns 0.0: common-neighbor confirmations are sent in iteration 2 but only reach vertex A in iteration 3, which the `process()` method does not handle. E.g. `jaccard_similarity(4, 6)` returns 0.0 although the mathematical value is 1/3. The existing (1,3) case passes only because 1 and 3 happen to be adjacent. This seems worth a separate issue; happy to file one if the maintainers agree.

Louvain / ASSP were intentionally not added: their tie-breaking depends on HashMap iteration order, which makes hand-verified expectations impractical.


geaflow/geaflow-dsl/geaflow-dsl-runtime/src/test/resources/expect/gql_algorithm_closeness_centrality_modern.txt
```diff
@@ -0,0 +1 @@
+1,1.0
```
Leomrlin, line 1: This expectation appears to lock in incorrect behavior for a disconnected directed graph. From vertex 1, only vertices 2, 3, 4, and 5 are reachable; vertex 6 is unreachable, and the sum of the finite distances is 1 + 1 + 1 + 2 = 5.

  The current implementation counts all five other vertices in the numerator while omitting the unreachable distance from the denominator, which produces 1.0 and allows an unreachable vertex to inflate the centrality score.

  Could we first define the intended semantics for disconnected graphs and then update both the implementation and this expectation? For example, Wasserman-Faust normalization would give (4 / 5) * (4 / 5) = 0.64 here, while the classical definition with an infinite distance would give 0.

geaflow/geaflow-dsl/geaflow-dsl-runtime/src/test/resources/query/gql_algorithm_jaccard_similarity_002.sql
```diff
@@ -0,0 +1,34 @@
...
+   jaccard_coefficient double
+) WITH (
+      type='file',
+      geaflow.dsl.file.path='${target}'
+);
+
+USE GRAPH modern;
+
+INSERT INTO result_tb
+CALL jaccard_similarity(1, 4) YIELD (vertex_a, vertex_b, jaccard_coefficient)
```
Leomrlin, line 32: This case currently passes for the wrong reason and does not exercise the known non-adjacent-vertex path.

  Because vertices 1 and 4 are directly adjacent, vertex 1 receives vertex 4's type-0 inquiry in iteration 2 and counts it as a common-neighbor confirmation. The actual common neighbor, vertex 3, sends its confirmation during iteration 2, so that message only arrives in iteration 3; however, JaccardSimilarity has no iteration-3 handler.

  The incorrect intersection count is also 1, so the implementation happens to return the mathematically correct value 0.2. The existing (1, 3) test has the same adjacent-vertices/one-common-neighbor shape.

  Please use a discriminating case such as (4, 6), whose expected result is 1 / 3, and fix the iteration-3 aggregation. If fixing the algorithm is outside this PR's scope, I suggest removing this case and tracking the bug in a separate linked issue.

PR #832, opened 2026-08-13
[ISSUE-830] Fix inverted embedding value filter that dropped digit-free text

### LINKED ISSUE #830: geaflow-ai: is the embedding value filter meant to drop prose and keep digits?
**Describe the bug**

In `geaflow-ai`, the value filter that feeds `EmbeddingIndexStore` keeps values containing a digit or one of `. _ - @ + ! $ % & = ~`, and drops everything else. The practical effect is that ordinary prose is never embedded, while a bare id, a date or a run of punctuation is.

`SearchUtils.isAllAllowedChars` returns on the first character **inside** the character set rather than the first one outside it:

```java
for (char c : str.toCharArray()) {
    if (IGNORE_CHARS.contains(c)) {
        return false;
    }
}
return true;
```

and `SubgraphSemanticPromptFunction.verbalize(GraphEntity)`, the overload `EmbeddingIndexStore` feeds on, keeps a value when the predicate is false:

```java
.filter(str -> !SearchUtils.isAllAllowedChars(str))
```

**Before opening this as a bug I want to ask whether it is intended**, because I cannot rule out a deliberate "only vectorize identifier-like values" policy. My reading is that it is not intended, for four reasons that are independent of each other. If any of them is a misreading, please say so and I will close this.

1. **The name, the Javadoc and the call site all agree with the opposite behaviour; only the loop body disagrees.** The method is named "is *all* allowed chars" and documented as `@return true if all characters are allowed`. The set contains digits and symbols but no letters, so under the documented reading `"abc"` yields false and is kept, `"12345"` yields true and is dropped. That is the behaviour the negation at the call site asks for. One place disagreeing with three seems more likely than three places disagreeing with one.
2. **An empty value is sent to the embedding model.** The predicate returns false for null and empty, and the call site keeps it. The early return even carries the comment `Consider empty/null invalid`, so `false` was meant as "invalid", yet the caller keeps precisely the invalid ones.
3. **The keyword path applies no such filter.** Keyword indexing goes through `verbalize(SubGraph)`, which renders everything. If dropping prose here were a division of labour between the two retrieval paths, the two would have to be complementary; instead one takes everything and the other only values containing a digit.
4. **`formatQuery` runs immediately after the filter and strips much of what the filter selected on.** `EXCLUDED_CHARS` and the ignorable set overlap on `-` and `.`, so a value kept because it contains `-` has that `-` replaced by a space in the next step.

**Expected behavior**

A value should be skipped when it carries no meaning on its own, such as a bare id or a date, and kept otherwise. Whether a value is embeddable should not depend on it containing a digit.

**Additional context**

*Reproduction on `master` (`6312f97d`), no network and no API key needed.* Two single vertex graphs whose text differs only by an appended digit, `EmbeddingIndexStore.initStore` with `new ModelConfig(null, null, null, null)`:

| vertex text | index lines written | outcome |
|---|---|---|
| `no digits here at all` | 0 | completes, no request attempted |
| `has a digit 7 in it` | 0 | attempts the request, fails on the null url |

The digit-free case is silent in a second way: it logs `Successfully added 1 new index items. Total indexed: 1` while having produced nothing. The entity is registered in `indexStoreMap` with an empty vector list, so it can never be recalled, and nothing distinguishes that from a successful build.

*Why this has not been noticed.* The only test covering embeddings, `GraphMemoryTest#testLdbcMainPipeline`, uses LDBC data whose values mix letters and digits. Both the current and the corrected predicate keep those, so the defect is invisible there. It also means the module's own end-to-end scenario is affected: `MemoryServerTest` imports 532 prose chunks, none of which would ever be embedded.

*Proposed fix.* I would split it in two, because only one half depends on the answer to the question above.

Unconditional, regardless of the intended semantics:

- Warn when an entity yields no embeddable text, with a count and an example key, and make the summary line report how many entities actually hold vectors rather than how many were queued. Silently reporting success for a no-op is a defect under any reading.

Conditional on the semantics being a bug:

- Correct the predicate to return true only when every character is ignorable, and treat null and empty as ignorable, since callers use it to decide what to skip and there is nothing in them to index.
- Rename it to `isAllIgnorableChars`. Renaming rather than correcting in place is deliberate: with the name unchanged, any caller that depended on the previous meaning would silently flip behaviour, whereas a rename makes it fail to compile. Both call sites are in `SubgraphSemanticPromptFunction` and keep reading `!isAllIgnorableChars(value)`.

*Blast radius, measured before changing anything.* On the LDBC dataset, of 168 entities the set wanting vectors is identical before and after the correction, and none of them are absent from the committed `LDBCEmbeddingIndexStore` file. The two predicates differ only on values that are entirely digits or punctuation, and on values that are entirely letters; LDBC entities have neither. So the committed embedding index needs no regeneration and `GraphMemoryTest` passes unchanged.

I have the change and tests ready locally, 7 cases covering the predicate, the property that appending a digit must not change whether a value is indexable, verbalization of digit-free vertex and edge text, and the store either attempting a request or correctly declining to. All 7 fail against the current behaviour. Happy to open a PR once the intent is confirmed, or to reduce it to the unconditional half if the filter is meant to work the way it does.

E2ern1ty:
Fixes #830.

The value filter feeding `EmbeddingIndexStore` kept values containing a digit or one of `. _ - @ + ! $ % & = ~` and dropped everything else, so ordinary prose was never embedded while a bare id, a date or a run of punctuation was. `SearchUtils.isAllAllowedChars` returned on the first character **inside** the ignorable set rather than the first one outside it, which is the negation of both its name and its Javadoc, and `SubgraphSemanticPromptFunction.verbalize(GraphEntity)` keeps a value when that predicate is false.

Two changes, matching the split agreed in the issue.

**Correct the predicate, and rename it to `isAllIgnorableChars`.** It now returns true only when every character is ignorable, and treats null and empty as ignorable since callers use it to decide what to skip and there is nothing in them to index. The rename is deliberate rather than a fix in place: with the name unchanged, a caller depending on the previous meaning would silently flip behaviour, whereas a rename makes it fail to compile. Both call sites are in `SubgraphSemanticPromptFunction` and read the same as before, `!isAllIgnorableChars(value)`. `IGNORE_CHARS` becomes `IGNORABLE_CHARS`, and the Javadoc now describes the set as characters that carry no meaning on their own.

**Stop reporting entities with no embeddable text as indexed.** This is the half that held regardless of the predicate's intended semantics, and it was requested on the issue. A digit-free entity logged `Successfully added 1 new index items. Total indexed: 1` while producing nothing: the entity was registered in `indexStoreMap` with an empty vector list, so it could never be recalled, and nothing distinguished that from a successful build. `indexBatch` now warns with a count and an example key when entities yield no text, and the summary line reports how many entities actually hold vectors rather than how many were queued. From review, `getEntityIndex` also logs at debug level when an entity was checked and held nothing, which previously looked identical to an entity nobody had looked at.

### How was this PR tested?

**Unit tests.** `IgnorableTextFilterTest`, 7 cases: the predicate over values that do and do not carry meaning, the property that whether a value is kept must not depend on it containing a digit, verbalization of digit-free vertex and edge text, verbalization dropping a value that is only a date, and the store either attempting a request or correctly declining to. All 7 fail against the previous behaviour.

**Index lifecycle over successive runs.** `EmbeddingIndexLifecycleTest`, added from review, runs three `initStore` calls against a local `/v1/embeddings` endpoint on an ephemeral port, so request and response serialization, the on-disk index file and the incremental decision are all real while the suite stays offline and deterministic. A meaningful vertex and edge are embedded and persisted while a date-only vertex is not, 2 lines; an unchanged graph on a second run issues no request and restores its vectors from the file; and when the date is replaced by meaningful text, that entity alone is embedded, the request body carries only its text, and the file grows to 3 lines. The last phase is the regression test for the review question of whether an entity that once yielded nothing is excluded for good: it is not, because `initStore` rebuilds `indexStoreMap` from the index file alone and nothing is persisted for such an entity.

Full `geaflow-ai` suite passes, 14 tests, 0 checkstyle violations.

**End-to-end against a real embedding service.** Both `master` (`3f73eb55`) and this branch were run through the full production path against SiliconFlow `BAAI/bge-m3`: real HTTPS requests, real 1024-dimension vectors, real on-disk index file, real recall through `GraphMemoryServer`. Corpus is the module's own `text/Confucius`, 532 chunks of Chinese prose, of which 532 contain no digit. Same harness source compiled against both trees, so the numbers are comparable.

| | master | this branch |
|---|---|---|
| entities holding vectors | 0 of 532 | 532 of 532 |
| index file lines written | 0 | 548 |
| embedding requests issued | 0 | 23, for 549 texts |
| summary log | `Successfully added 532 new index items. Total indexed: 532` | `Successfully added 532 new index items. Entities holding vectors: 532 of 532` |
| recall of a known chunk, embedding vector only | not possible, no vectors | returned, cosine 0.9999 |
| recall for an unseen question, `学习和思考的关系是什么？` | not possible | top chunk is `子曰：“学而时习之，不亦说乎？…”`, cosine 0.5835 |
| rebuild from the index file just written | n/a | 0 new requests, 532 of 532 restored, recall still returns the chunk |

So on `master` the whole corpus was reported as indexed while nothing was embedded and no request was ever issued. Recall was exercised with an embedding vector only, no keyword vector and no other index store registered, so anything returned had to come from the embedding index. The written index file holds 548 records of 1024 real dimensions each, 11.35 MB, keyed by entity, with no zero vectors.

The counterpart case on the same endpoint, 20 date-only values, which is what the old predicate selected *for*:

| | master | this branch |
|---|---|---|
| entities holding vectors | 20 of 20 | 0 of 20 |
| embedding requests issued | 1, for 20 inputs | 0 |
| log | `Successfully added 20 new index items. Total indexed: 20` | `WARN 20 of 20 entities have no embeddable text and will hold no vectors, for example Vd1chunk`, then `Entities holding vectors: 0 of 20` |

The two tables together are the inversion the issue describes, measured against a live model rather than argued: `master` spends requests on dates and none on prose, this branch does the opposite and says so when there is nothing to embed.

**Blast radius, measured before changing anything.** On the LDBC test dataset, of 168 entities the set wanting vectors is identical before and after the correction, and none of them are absent from the committed `LDBCEmbeddingIndexStore` file. The two predicates differ only on values that are entirely digits or punctuation, and on values that are entirely letters; LDBC entities have neither. The committed embedding index therefore needs no regeneration and `GraphMemoryTest` passes unchanged, which is also why the defect was invisible until now. It does mean the module's own end-to-end scenario was affected: `MemoryServerTest` imports those same 532 prose chunks, none of which would previously have been embedded.

**Known gap, not addressed here, now #844.** An entity already present in the index file whose value later changes is not re-embedded, because the index key is `ModelUtils.getGraphEntityKey`, id and label with no content in it, so the vector for the old value keeps matching while the current value has no vector at all. Reproduced on `master` while writing that issue up: `cosine(stored, old value) = 1.0000` against `cosine(stored, current value) = 0.2357`, a query for the old value returns the entity and a query for the current value does not. It predates this change, which only altered which values count as embeddable, and it needs a content hash in the record plus an invalidation path.

### DIFF
--- a/geaflow-ai/src/main/java/org/apache/geaflow/ai/index/EmbeddingIndexStore.java
+++ b/geaflow-ai/src/main/java/org/apache/geaflow/ai/index/EmbeddingIndexStore.java
@@ -262,15 +262,24 @@ public class EmbeddingIndexStore implements IndexStore {
 
     @Override
     public List<IVector> getEntityIndex(GraphEntity entity) {
-        if (entity != null && indexStoreMap.get(entity) != null) {
-            List<EmbeddingService.EmbeddingResult> resultList = indexStoreMap.get(entity);
-            List<IVector> result = new ArrayList<>();
-            for (EmbeddingService.EmbeddingResult res : resultList) {
-                double[] embedding = res.embedding;
-                result.add(new EmbeddingVector(embedding));
-            }
-            return result;
+        if (entity == null) {
+            return Collections.emptyList();
         }
-        return Collections.emptyList();
+        List<EmbeddingService.EmbeddingResult> resultList = indexStoreMap.get(entity);
+        if (resultList == null) {
+            return Collections.emptyList();
+        }
+        if (resultList.isEmpty()) {
+            // An entity that was looked at but held no embeddable text is indistinguishable from an
+            // entity nobody has looked at yet, since both give an empty result. Say which it was.
+            LOGGER.debug("Entity {} was checked and holds no embeddable text",
+                    ModelUtils.getGraphEntityKey(entity));
+            return Collections.emptyList();
+        }
+        List<IVector> result = new ArrayList<>(resultList.size());
+        for (EmbeddingService.EmbeddingResult res : resultList) {
+            result.add(new EmbeddingVector(res.embedding));
+        }
+        return result;
     }
 }
--- a/geaflow-ai/src/main/java/org/apache/geaflow/ai/index/EmbeddingIndexStore.java
+++ b/geaflow-ai/src/main/java/org/apache/geaflow/ai/index/EmbeddingIndexStore.java
@@ -172,8 +172,16 @@ public class EmbeddingIndexStore implements IndexStore {
             batchEntitiesBuffer.clear();
         }
 
-        LOGGER.info("Successfully added {} new index items. Total indexed: {}",
-                addedCount, indexStoreMap.size());
+        // Count entries that actually carry vectors, not entities that were queued: an entity with
+        // no embeddable text is registered with an empty list and must not be reported as indexed.
+        long withVectors = 0;
+        for (List<EmbeddingService.EmbeddingResult> vectors : indexStoreMap.values()) {
+            if (vectors != null && !vectors.isEmpty()) {
+                withVectors++;
+            }
+        }
+        LOGGER.info("Successfully added {} new index items. Entities holding vectors: {} of {}",
+                addedCount, withVectors, indexStoreMap.size());
     }
 
     private List<String> indexBatch(EmbeddingService service, List<GraphEntity> pendingEntities) {
@@ -182,13 +190,24 @@ public class EmbeddingIndexStore implements IndexStore {
         }
         List<String> pendingTexts = new ArrayList<>(pendingEntities.size());
         Map<GraphEntity, Pair<Integer, Integer>> entity2StartEndPair = new HashMap<>();
+        List<GraphEntity> withoutText = new ArrayList<>();
         for (GraphEntity e : pendingEntities) {
             Integer start = pendingTexts.size();
             pendingTexts.addAll(ModelUtils.splitLongText(
                 Constants.EMBEDDING_INDEX_STORE_SPLIT_TEXT_CHUNK_SIZE,
                     verbFunc.verbalize(e).toArray(new String[0])));
             Integer end = pendingTexts.size();
+            if (start.equals(end)) {
+                withoutText.add(e);
+            }
             entity2StartEndPair.put(e, Pair.of(start, end));
+        }
+        if (!withoutText.isEmpty()) {
+            // Say so rather than reporting these as indexed. An entity whose values are all
+            // ignorable yields no text, so it gets an empty vector list and can never be recalled.
+            LOGGER.warn("{} of {} entities have no embeddable text and will hold no vectors, "
+                    + "for example {}", withoutText.size(), pendingEntities.size(),
+                ModelUtils.getGraphEntityKey(withoutText.get(0)));
         }
 
         Gson gson = new Gson();
--- a/geaflow-ai/src/main/java/org/apache/geaflow/ai/operator/SearchUtils.java
+++ b/geaflow-ai/src/main/java/org/apache/geaflow/ai/operator/SearchUtils.java
@@ -31,17 +31,17 @@ public class SearchUtils {
             '*', '#', '-', '?', '`', '{', '}', '[', ']', '(', ')', '>', '<', ':', '/', '.'
     ));
 
-    // Set of allowed characters for validation in isAllAllowedChars
-    // Includes: digits (0-9), and some common safe symbols
-    private static final Set<Character> IGNORE_CHARS = buildIgnoredChars();
+    // Characters that carry no meaning on their own: digits and a few common symbols.
+    // A value made up entirely of these is not worth indexing or embedding.
+    private static final Set<Character> IGNORABLE_CHARS = buildIgnorableChars();
 
     /**
-     * Builds the set of allowed characters for input validation.
-     * Includes alphanumeric characters and selected common symbols.
+     * Builds the set of characters that carry no meaning on their own.
+     * Digits and a few common symbols.
      *
-     * @return an unmodifiable set of ignored characters
+     * @return an unmodifiable set of ignorable characters
      */
-    private static Set<Character> buildIgnoredChars() {
+    private static Set<Character> buildIgnorableChars() {
         Set<Character> ignored = new HashSet<>(32);
         // Add digits
         for (char c = '0'; c <= '9'; c++) {
@@ -87,18 +87,30 @@ public class SearchUtils {
     }
 
     /**
-     * Checks whether all characters in the given string are within the allowed character set.
-     * Useful for validating usernames, identifiers, or safe input formats.
+     * Whether the given value consists entirely of characters that carry no meaning on their own,
+     * and therefore has nothing worth indexing or embedding. A bare id, a date or a run of
+     * punctuation is ignorable; anything containing a letter or a CJK character is not.
      *
-     * @param str the string to validate
-     * @return true if all characters are allowed; false otherwise
+     * <p>Callers use this to skip values, so an empty or absent value is ignorable too: there is
+     * nothing in it to index.
+     *
+     * <p>This replaces {@code isAllAllowedChars}, whose loop returned on the first character
+     * <em>inside</em> the set rather than the first one outside it, making it the negation of both
+     * its own name and its own documentation. The practical effect was that ordinary prose was
+     * discarded while digit-only noise was kept, so an embedding store silently produced nothing
+     * for text that happened to contain no digit. The method is renamed rather than corrected in
+     * place, so that any caller depending on the previous meaning fails to compile instead of
+     * silently flipping behaviour.
+     *
+     * @param str the value to check
+     * @return true if every character is ignorable, or the value is null or empty
      */
-    public static boolean isAllAllowedChars(String str) {
+    public static boolean isAllIgnorableChars(String str) {
         if (str == null || str.isEmpty()) {
-            return false; // Consider empty/null invalid; adjust based on use case
+            return true;
         }
         for (char c : str.toCharArray()) {
-            if (IGNORE_CHARS.contains(c)) {
+            if (!IGNORABLE_CHARS.contains(c)) {
                 return false;
             }
         }
--- a/geaflow-ai/src/test/java/org/apache/geaflow/ai/operator/IgnorableTextFilterTest.java
+++ b/geaflow-ai/src/test/java/org/apache/geaflow/ai/operator/IgnorableTextFilterTest.java
@@ -1,0 +1,194 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.geaflow.ai.operator;
+
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import org.apache.geaflow.ai.common.config.Constants;
+import org.apache.geaflow.ai.common.model.ModelConfig;
+import org.apache.geaflow.ai.graph.GraphEntity;
+import org.apache.geaflow.ai.graph.LocalMemoryGraphAccessor;
+import org.apache.geaflow.ai.graph.io.Edge;
+import org.apache.geaflow.ai.graph.io.EdgeGroup;
+import org.apache.geaflow.ai.graph.io.EdgeSchema;
+import org.apache.geaflow.ai.graph.io.EntityGroup;
+import org.apache.geaflow.ai.graph.io.GraphSchema;
+import org.apache.geaflow.ai.graph.io.MemoryGraph;
+import org.apache.geaflow.ai.graph.io.Vertex;
+import org.apache.geaflow.ai.graph.io.VertexGroup;
+import org.apache.geaflow.ai.graph.io.VertexSchema;
+import org.apache.geaflow.ai.index.EmbeddingIndexStore;
+import org.apache.geaflow.ai.verbalization.SubgraphSemanticPromptFunction;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+
+/**
+ * The value filter feeding the embedding store must drop values that carry no meaning and keep the
+ * ones that do.
+ *
+ * <p>It used to do the opposite, because the predicate returned on the first character
+ * <em>inside</em> the ignorable set rather than the first one outside it. Ordinary prose was
+ * therefore discarded and digit-only noise kept, and an embedding store silently produced nothing
+ * for any text that happened to contain no digit: no request was issued, no error was raised, and
+ * the entity was still registered as indexed with an empty vector list.
+ */
+public class IgnorableTextFilterTest {
+
+    private static final String LABEL = "chunk";
+    private static final String EDGE_LABEL = "rel";
+
+    @Test
+    public void testValuesCarryingMeaningAreNotIgnorable() {
+        for (String value : new String[] {
+            "the alpaca grazes on the hillside",
+            "chunk 7 of the report",
+            "a",
+            "\u4e2d\u6587\u5185\u5bb9",
+            "id-42 belongs to Alice"}) {
+            Assertions.assertFalse(SearchUtils.isAllIgnorableChars(value),
+                "value carries meaning and must be kept: " + value);
+        }
+    }
+
+    @Test
+    public void testValuesWithoutMeaningAreIgnorable() {
+        for (String value : new String[] {
+            "12345", "2024-01-01", "----", "___", "42.0", "%%", null, ""}) {
+            Assertions.assertTrue(SearchUtils.isAllIgnorableChars(value),
+                "value carries no meaning and must be dropped: " + value);
+        }
+    }
+
+    /**
+     * The property that matters, stated without reference to the implementation: whether a value is
+     * kept must not depend on it containing a digit.
+     */
+    @Test
+    public void testKeepingAValueDoesNotDependOnContainingADigit() {
+        Assertions.assertEquals(
+            SearchUtils.isAllIgnorableChars("the alpaca grazes"),
+            SearchUtils.isAllIgnorableChars("the alpaca grazes 7"),
+            "adding a digit to prose must not change whether it is indexable");
+    }
+
+    @Test
+    public void testVerbalizationKeepsDigitFreeText() {
+        LocalMemoryGraphAccessor accessor = buildGraph("no digits here at all", "still none");
+        SubgraphSemanticPromptFunction func = new SubgraphSemanticPromptFunction(accessor);
+
+        GraphEntity vertex = accessor.getVertex(LABEL, "v1");
+        Assertions.assertEquals(Collections.singletonList("no digits here at all"),
+            func.verbalize(vertex),
+            "digit free vertex text must survive verbalization");
+
+        List<GraphEntity> edges = new ArrayList<>(accessor.getEdge(EDGE_LABEL, "v1", "v2"));
+        Assertions.assertEquals(1, edges.size());
+        Assertions.assertEquals(Collections.singletonList("plain edge text"),
+            func.verbalize(edges.get(0)),
+            "digit free edge text must survive verbalization");
+    }
+
+    @Test
+    public void testVerbalizationDropsValuesWithoutMeaning() {
+        LocalMemoryGraphAccessor accessor = buildGraph("2024-01-01", "still none");
+        SubgraphSemanticPromptFunction func = new SubgraphSemanticPromptFunction(accessor);
+        Assertions.assertEquals(Collections.emptyList(),
+            func.verbalize(accessor.getVertex(LABEL, "v1")),
+            "a value that is only a date carries nothing to embed");
+    }
+
+    /**
+     * End to end on the store, offline: an entity with digit free text must be queued for
+     * embedding. The unusable model config makes that observable without a service, since reaching
+     * the request at all fails on the null url.
+     */
+    @Test
+    public void testStoreRequestsEmbeddingsForDigitFreeText(@TempDir Path tempDir) {
+        LocalMemoryGraphAccessor accessor = buildGraph("no digits here at all", "still none");
+        Path indexFile = tempDir.resolve("index.jsonl");
+        EmbeddingIndexStore store = new EmbeddingIndexStore();
+
+        int retries = Constants.MODEL_CLIENT_RETRY_TIMES;
+        int interval = Constants.MODEL_CLIENT_RETRY_INTERVAL_MS;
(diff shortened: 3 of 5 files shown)


geaflow-ai/src/main/java/org/apache/geaflow/ai/operator/SearchUtils.java
```diff
@@ -87,18 +87,30 @@ public static String formatQuery(String query) {
...
+     * discarded while digit-only noise was kept, so an embedding store silently produced nothing
+     * for text that happened to contain no digit. The method is renamed rather than corrected in
+     * place, so that any caller depending on the previous meaning fails to compile instead of
+     * silently flipping behaviour.
+     *
+     * @param str the value to check
+     * @return true if every character is ignorable, or the value is null or empty
      */
-    public static boolean isAllAllowedChars(String str) {
+    public static boolean isAllIgnorableChars(String str) {
```
kitalkuyo-gita, line 108: First of all, thank you for your contribution. However, upon reviewing the repository, I noticed that there are still references to this function in some places. I have the following suggestions:

1. Perform a full-code search before merging: Search the entire `apache/geaflow` repository for all call sites of `isAllAllowedChars` to confirm that all instances have been updated in the PR. You should run a search like this:

```
grep -r "isAllAllowedChars" --include="*.java" .
```

2. Suggestion for a transition period regarding deprecated methods: Retain the old method but mark it as `@Deprecated`, and remove it only after 2–3 release cycles to avoid breaking downstream dependencies:

```
@Deprecated(since = "1.4.0", forRemoval = true)
public static boolean isAllAllowedChars(String str) {
return !isAllIgnorableChars(str);  // Delegate to the new method
}
```

geaflow-ai/src/main/java/org/apache/geaflow/ai/index/EmbeddingIndexStore.java
```diff
@@ -172,8 +172,16 @@ public void initStore(GraphAccessor graphAccessor, VerbalizationFunction func,
             batchEntitiesBuffer.clear();
         }
 
-        LOGGER.info("Successfully added {} new index items. Total indexed: {}",
-                addedCount, indexStoreMap.size());
+        // Count entries that actually carry vectors, not entities that were queued: an entity with
+        // no embeddable text is registered with an empty list and must not be reported as indexed.
+        long withVectors = 0;
+        for (List<EmbeddingService.EmbeddingResult> vectors : indexStoreMap.values()) {
```
kitalkuyo-gita, line 178: Insufficient Exception Handling in EmbeddingIndexStore

Risk Description: In the `indexBatch` method, when an entity lacks embeddable text, a warning log is generated, but the entity is still registered in `indexStoreMap` (with an empty list of warnings). This can lead to the following issues:

During a second initialization, the `containsKey` check skips these entities (assuming they are already indexed).
Queries return empty results instead of a clear signal indicating "not indexed."

```
// Inside the indexBatch method - lines 210-223
for (Map.Entry<GraphEntity, Pair<Integer, Integer>> entry : entity2StartEndPair.entrySet()) {
GraphEntity e = entry.getKey();
List<EmbeddingService.EmbeddingResult> embeddings = new ArrayList<>();
for (int i = entry.getValue().getLeft(); i < entry.getValue().getRight(); i++) {
// If start == end (entities in withoutText), this loop does not execute
if (StringUtils.isNotBlank(result.get(i))) {
// ...
embeddings.add(res);
}
}
// ⚠️ Registered regardless of whether embeddings is empty
indexStoreMap.put(e, embeddings);  // embeddings might be an empty list
}
```

Problem scenario: Assume the graph contains two entities:

1st run: Entity A has text (embedding successful); Entity B has no text (empty list registered).
2nd run: Entity B's value is updated to meaningful text.
`containsKey(B)` returns `true` → skipped.
Entity B is never re-embedded.
Recommendation:

1. Do not register entities that yield no embeddings, or use a separate set for "checked but empty" entities:

```
// New field
private Set<GraphEntity> checkedButEmpty = new HashSet<>();

// Modified indexBatch
for (Map.Entry<GraphEntity, Pair<Integer, Integer>> entry : entity2StartEndPair.entrySet()) {
GraphEntity e = entry.getKey();
List<EmbeddingService.EmbeddingResult> embeddings = new ArrayList<>();
for (int i = entry.getValue().getLeft(); i < entry.getValue().getRight(); i++) {
if (StringUtils.isNotBlank(result.get(i))) {
EmbeddingService.EmbeddingResult res = gson.fromJson(result.get(i),
EmbeddingService.EmbeddingResult.class);
res.input = ModelUtils.getGraphEntityKey(e);
formatResult.add(gson.toJson(res)); 
embeddings.add(res);
}
}
if (!embeddings.isEmpty()) {
indexStoreMap.put(e, embeddings);
} else {
checkedButEmpty.add(e);  // Explicitly mark as checked but containing no content
}
}

// Modify the check logic in initStore
if (!indexStoreMap.containsKey(vertex) && !checkedButEmpty.contains(vertex)
&& !batchEntitiesBuffer.contains(vertex)) {
// Only entities that have truly not been processed enter the batch
batchEntitiesBuffer.add(vertex);
pendingEntities.add(vertex);
}
```

2. Enhance the distinction of return values ​​for getEntityIndex:

```
@Override
public List<IVector> getEntityIndex(GraphEntity entity) {
if (entity != null) {
List<EmbeddingService.EmbeddingResult> resultList = indexStoreMap.get(entity);
if (resultList != null) {
if (resultList.isEmpty()) {
LOGGER.debug("Entity {} was checked but contains no embeddable text", entity);
} else {
List<IVector> result = new ArrayList<>();
for (EmbeddingService.EmbeddingResult res : resultList) {
result.add(new EmbeddingVector(res.embedding));
}
return result;
}
}
}
return Collections.emptyList();
}

File: geaflow-ai/src/test/java/org/apache/geaflow/ai/operator/IgnorableTextFilterTest.java [hunk 3]
```diff
@@ -0,0 +1,194 @@
...
+ * for any text that happened to contain no digit: no request was issued, no error was raised, and
+ * the entity was still registered as indexed with an empty vector list.
+ */
+public class IgnorableTextFilterTest {
+
+    private static final String LABEL = "chunk";
+    private static final String EDGE_LABEL = "rel";
+
+    @Test
+    public void testValuesCarryingMeaningAreNotIgnorable() {
```
kitalkuyo-gita, line 65: Although the newly added `IgnorableTextFilterTest` contains seven test cases, it lacks a crucial incremental indexing scenario:

Here is a simulated scenario:
```
// All tests in IgnorableTextFilterTest follow this pattern:
// 1. Build a new graph
// 2. Perform the initial initStore
//
// There are no tests covering:
// 1. Graph modification (an entity's text changes from "meaningless" to "meaningful")
// 2. Whether initStore correctly re-indexes upon a subsequent run
Missing test case:
```

Missing capability:
```
/**
* Missing test: Behavior when entity text changes from meaningless to meaningful
*/
@Test
public void testReindexWhenEntityValueChangesFromIgnorableToMeaningful(@TempDir Path tempDir)
throws Exception {
// First run: Entity A's value is "2024-01-01" (meaningless)
LocalMemoryGraphAccessor accessor1 = buildGraph("2024-01-01", "still none");
EmbeddingIndexStore store = new EmbeddingIndexStore();
store.initStore(accessor1, new SubgraphSemanticPromptFunction(accessor1),
tempDir.resolve("index1.jsonl").toString(),
new ModelConfig(null, null, null, null));

// Verify that no requests were sent during the first run
Assertions.assertTrue(Files.readAllLines(tempDir.resolve("index1.jsonl")).isEmpty());

// Second run: The value of the same entity is modified to meaningful text
//  Question: How do we simulate this scenario? Entity A already has a record in `indexStoreMap` (an empty vector list)
// `containsKey` returns true, causing reprocessing to be skipped
LocalMemoryGraphAccessor accessor2 = buildGraph("meaningful text", "still none");
// ... This scenario cannot be tested because the entity objects will differ
}
```

Suggestion:

Add a top-level integration test:

```
@Test
public void testCompleteIndexLifecycle(@TempDir Path tempDir) throws Exception {
// Phase 1: Initial indexing
LocalMemoryGraphAccessor graph1 = buildGraphWithMixedContent();
EmbeddingIndexStore store1 = new EmbeddingIndexStore();
store1.initStore(graph1, new SubgraphSemanticPromptFunction(graph1),
tempDir.resolve("index.jsonl").toString(), getRealModelConfig());
int firstRound = Files.readAllLines(tempDir.resolve("index.jsonl")).size();

// Phase 2: Delete some entities, add new entities
LocalMemoryGraphAccessor graph2 = buildGraphWithModifiedContent();
EmbeddingIndexStore store2 = new EmbeddingIndexStore();
store2.initStore(graph2, new SubgraphSemanticPromptFunction(graph2),
tempDir.resolve("index.jsonl").toString(), getRealModelConfig());
int secondRound = Files.readAllLines(tempDir.resolve("index.jsonl")).size();

// Verify the correctness of incremental indexing
Assertions.assertTrue(secondRound > firstRound || secondRound == firstRound);
}
```

geaflow-ai/src/main/java/org/apache/geaflow/ai/operator/SearchUtils.java
E2ern1ty: Thanks. I ran exactly the search you asked for, and there are no remaining call sites.

```
$ grep -r "isAllAllowedChars" --include="*.java" .
./geaflow-ai/src/main/java/org/apache/geaflow/ai/operator/SearchUtils.java:97:
     * <p>This replaces {@code isAllAllowedChars}, whose loop returned on the first character
```

One hit, and it is the Javadoc line in this file explaining what the method replaces. Widening the search to every file type gives two more, both generated build output rather than source: `target/site/jacoco/.../SearchUtils.java.html` and `target/apidocs/.../SearchUtils.html`. Those are rebuilt from this Javadoc. If the references you saw were in a `target/` directory or in the doc text, that is what they were. The two real call sites, both in `SubgraphSemanticPromptFunction`, are updated in this PR and now read `!isAllIgnorableChars(value)`.

On the deprecated shim, I would rather not add it, for three reasons.

The delegation is not behaviour preserving, which is the part that worries me most. `isAllAllowedChars` returned false as soon as it met *one* ignorable character, whereas `isAllIgnorableChars` returns true only when *every* character is ignorable. Those are not complements. Take `"a1"`: the old method returned false, because `1` is in the set, while `!isAllIgnorableChars("a1")` is `!false`, that is true. So `return !isAllIgnorableChars(str)` would silently flip the answer for any mixed value, and mixed values are the common case in real data. LDBC entity values are exactly letters plus digits. A shim that quietly changes behaviour for the majority of inputs is the hazard the rename was meant to remove, not a mitigation of it.

It also does not compile here. `since` and `forRemoval` were added to `@Deprecated` in Java 9, and this project builds at `<jdk.version>1.8</jdk.version>` (root `pom.xml`). And `1.4.0` is not a GeaFlow version; the tree is at `0.8.0-SNAPSHOT`.

There is no downstream contract to protect either. `SearchUtils` is not present in any release tag, `v0.8.0-rc1` included:

```
$ for t in v0.8.0-rc1 v0.7.0 v0.7.0-rc3 v0.7.0-rc2 v0.7.0-rc1; do
    printf "%s: " $t; git cat-file -e "$t:geaflow-ai/.../operator/SearchUtils.java" 2>/dev/null \
      && echo present || echo absent; done
v0.8.0-rc1: absent
v0.7.0: absent
...
```

It arrived with #716 in January, after the rc, so no published artifact has ever exposed this method and nothing outside the repo can be calling it. A deprecation window over two or three release cycles would keep a method whose semantics are the defect this PR fixes, and it would protect no one.

If you would still like a transition period, the only shim I would be comfortable adding is one that preserves the original contract rather than approximating it, `return containsAnyIgnorableChar(str)` under a name that says what it does. Happy to add that if you want it, but given that the class is unreleased I think deleting the old name is the better trade. Your call.

geaflow-ai/src/main/java/org/apache/geaflow/ai/index/EmbeddingIndexStore.java
E2ern1ty: Good thing to check, and I went looking for it rather than reasoning about it. The skip you describe does not happen, and I have added a test that would fail if it did: `EmbeddingIndexLifecycleTest`, in `b007b1a4`.

The reason is in `initStore` rather than in `indexBatch`. Every call begins with

```java
this.indexStoreMap = new HashMap<>();
```

and then repopulates the map from the index file only. Nothing is ever written to that file for an entity with no embeddable text, because `indexBatch` returns records for embedded texts alone. So on the next run the empty entry from the previous run does not exist, `containsKey` is false, and the entity is queued again. Your scenario ends the other way round: entity B does get re-embedded as soon as its value carries meaning.

The test runs three `initStore` calls against a local `/v1/embeddings` endpoint, so the request and response serialisation, the file and the decision about what to embed are all real, with no model service and no key. Run 1: a meaningful vertex and a meaningful edge are embedded and persisted, a date-only vertex is not, 2 lines in the file. Run 2, unchanged graph: 0 requests, vectors restored from the file, still 2 lines. Run 3, the date replaced by `the master replied in the temple`: exactly 1 request, its body contains the new text and does not contain the already-indexed text, the file grows to 3 lines, and `getEntityIndex` for that entity is no longer empty. The log from run 2 shows the mechanism directly, `1 of 1 entities have no embeddable text`, the 1 being that entity queued again rather than skipped.

On not registering the empty list at all: I looked at that and it would cost something. `MemoryGraph.scanEdge(vertex)` returns both out-edges and in-edges, so `initStore` reaches the same edge from each of its endpoints. For an edge with no embeddable text, the empty entry in `indexStoreMap` is what makes the second visit a no-op. Dropping it without a replacement means verbalizing that edge twice per run, and the `checkedButEmpty` set restores the skip but also adds a second piece of state that has to be kept in step with the map and reset on every `initStore`, for a saving of one local `verbalize` call. Given that the cross-run recovery above depends on those entities *not* being remembered, I would rather keep one map whose meaning is "these are the vectors I have" than two structures that have to agree.

Your second suggestion I have taken, in a smaller form. `getEntityIndex` returned an empty list for two different situations, an entity that was checked and held nothing, and an entity nobody has looked at. It still returns the same value, since a caller wants vectors either way and has nothing useful to do with the distinction, but the first case now logs at debug level instead of being silent.

One nearby gap that is real and that I have deliberately not touched: an entity that is *already* in the index file and whose value then changes is not re-embedded, because the index key is `ModelUtils.getGraphEntityKey`, which is id and label with no content in it, so the stale vector keeps matching. That is independent of this fix, it predates it, and fixing it means a content hash in the record and an invalidation path. Happy to open a separate issue for it if you agree it is worth one.

geaflow-ai/src/test/java/org/apache/geaflow/ai/operator/IgnorableTextFilterTest.java
E2ern1ty: Agreed, that gap was real. Added as `EmbeddingIndexLifecycleTest` in `b007b1a4`, and it covers the scenario your snippet was reaching for.

Two departures from the sketch. It does not need `getRealModelConfig()`: the test starts a local `/v1/embeddings` endpoint on an ephemeral port and returns deterministic vectors, so request and response serialisation, the on-disk index file and the incremental decision are all exercised for real, while the suite stays offline, needs no key and stays deterministic. And the assertion is exact rather than `secondRound > firstRound || secondRound == firstRound`, which holds for any outcome and so cannot fail: the test pins the line count to 2, then 2, then 3, and checks which text the request carried.

You noted in the comment that the change-of-value case looked untestable because the entity objects differ between the two graphs. That turns out not to be the obstacle, because the store does not match entities by object identity across runs. It writes and reads records keyed by `ModelUtils.getGraphEntityKey`, id and label, and `initStore` rebuilds its map from the file through `key2EntityMap`. So rebuilding the graph with the same ids and different text is exactly how the scenario is set up: `buildGraph(MEANINGFUL, IGNORABLE)` for the first two runs, then `buildGraph(MEANINGFUL, "the master replied in the temple")` for the third.

Phases, and what each pins down:

1. Mixed content. The meaningful vertex and the meaningful edge are embedded in one request and persisted, 2 lines; the date-only vertex holds no vector and contributes no line.
2. Same graph, new store, same file. 0 requests, vectors restored from the file, no duplicate records appended.
3. The date is replaced with meaningful text. Exactly 1 request, whose body contains the new text and not the text already in the file; the file grows to 3 lines; that entity now holds a vector.

Phase 3 is also the regression test for the concern in your other comment, that an entity yielding nothing once would be excluded for good. Full suite is now 14 tests, 0 checkstyle violations.

geaflow-ai/src/main/java/org/apache/geaflow/ai/operator/SearchUtils.java
kitalkuyo-gita: Okay, thank you very much for checking again. I've looked it over carefully, and indeed most of the references are documentation comments. You can delete the old names.

geaflow-ai/src/main/java/org/apache/geaflow/ai/index/EmbeddingIndexStore.java
kitalkuyo-gita: There is indeed a vulnerability nearby that I deliberately avoided: entities already existing in the index file using `ModelUtils.getGraphEntityKey` will not be re-embedded after their values ​​change because the index key is an empty `id` and `tag`, so outdated vectors still match. This is unrelated to this fix; it predates it, and fixing it would require adding a content hash and an invalidation path to the record. If you think it's worth creating a separate issue for this, I'd be happy to do so.

Regarding this, which could potentially cause fact-checking issues when old and new facts are switched, you're welcome to open a new issue so we can discuss this.

E2ern1ty: Opened as #844. Before writing it up I reproduced it, and it is worse than a stale vector: the values get swapped in both directions.

One vertex, value changed between two runs against the same index file, on `master` (`3f73eb55`), with a local endpoint that records what it is asked to embed:

| | requests | index lines | vectors held |
|---|---|---|---|
| run 1, `the master said learning without thought is labour lost` | 1 | 1 | 1 |
| run 2, same id, `the stable burned down and nobody asked about the horse` | 0 | 1 | 1 |

The vector the entity holds after run 2 belongs to the value that is no longer there, `cosine(stored, old value) = 1.0000` against `cosine(stored, current value) = 0.2357`. Recall then goes the wrong way on both queries: a query for the old value, which is not in the graph any more, returns the entity, while a query for the current value does not. And because the subgraph is verbalised from the current graph, the first case hands the model a context containing the *new* value while having matched on the *old* one. That is your fact-switching concern, and I think it is the part that matters most: the retrieval reason and the returned content disagree, so nothing downstream can tell that it happened.

Small correction to the paraphrase, in case it matters for the issue: the key is not empty, it is `PREFIX_V + id + label`, for example `Vv1chunk`. The problem is not that it is empty but that nothing in it derives from the embedded text, so a value change leaves the key identical and `initStore` reads that as already indexed.

One more thing the same run turned up, which I put in #844 as a secondary observation: a record whose entity has been deleted from the graph is never pruned. It is dropped during the rebuild, since `key2EntityMap` has no such key, but the line stays in the file, so the file only grows, and a later entity reusing that id would silently adopt the old vector.

Directions are sketched in the issue rather than settled: a hash of the embedded text in each record, compared during the rebuild, with a mismatch counting as not indexed. That changes the file format, so it needs a decision on how to treat existing index files, and pruning wants a rewrite rather than an append, which is probably separate compaction work. Happy to pick it up once you have a view on the record format. Keeping it out of #832, which only changed which values count as embeddable.

geaflow-ai/src/main/java/org/apache/geaflow/ai/operator/SearchUtils.java
E2ern1ty: Thanks for taking another look. Leaving the rename as it stands then, with the old name gone.

kaori-seasons (approved):
LGTM

kitalkuyo-gita (approved):
LGTM

Loognqiang (approved):
LGTM

PR #858, opened 2026-08-26
[ISSUE-844] fix(ai): invalidate an embedding vector when its text changes

### LINKED ISSUE #844: geaflow-ai: embedding vectors are never invalidated when an entity's value changes
**Describe the bug**

`EmbeddingIndexStore` keys each persisted vector by `ModelUtils.getGraphEntityKey`, which is a prefix, the id and the label, with nothing derived from the value that was embedded:

```java
return Constants.PREFIX_V + ((GraphVertex) entity).getVertex().getId() + entity.getLabel();
```

`initStore` treats the presence of that key in the index file as "this entity is indexed" and skips it. So when an entity keeps its id but its value changes, the vector for the old value stays and is never refreshed. Nothing detects it: no request is issued, no warning is logged, and the record on disk is unchanged.

The consequence is not only a stale vector, it is a swap in both directions. Retrieval for the old value still reaches the entity, while the entity's current value has no vector anywhere and cannot be reached at all. Since the subgraph handed to the model is verbalized from the *current* graph, a query about the old value returns a context containing the new value. This is the "old and new facts switched" case raised in [#832 review](https://github.com/apache/geaflow/pull/832#discussion_r3828997114): the reason a chunk was retrieved and the content the model is then shown disagree, which is exactly the situation a fact-checking or provenance layer cannot detect from the outside.

Opened at the reviewer's suggestion on #832. This predates and is independent of that PR, which only changed which values are considered embeddable.

**Expected behavior**

A vector should be invalidated when the text it was produced from changes. An entity whose value changed should be re-embedded on the next `initStore`, and retrieval should reflect the current value rather than a previous one.

**Additional context**

*Reproduction on `master` (`3f73eb55`), no key and no network needed, using a local OpenAI-compatible `/v1/embeddings` endpoint that records what it is asked to embed.* One vertex `v1`, value changed between two runs, same index file:

| | requests issued | index file lines | vectors held by `v1` |
|---|---|---|---|
| run 1, value `the master said learning without thought is labour lost` | 1 | 1 | 1 |
| run 2, same id, value `the stable burned down and nobody asked about the horse` | 0 | 1 | 1 |

The vector `v1` holds after run 2 is the one produced from the value that is no longer there:

```
cosine(stored vector, embedding of the old value) = 1.0000
cosine(stored vector, embedding of the current value) = 0.2357
```

and recall through `GraphMemoryServer`, with an embedding vector only and no other index store registered, goes the wrong way on both queries:

| query | entity returned |
|---|---|
| the old value, which is no longer in the graph | yes |
| the current value | no |

The first row is a false positive whose verbalized context shows the current value; the second is the current value being unreachable.

*Secondary observation from the same run.* A record whose entity has been deleted from the graph is never pruned. `initStore` drops it while rebuilding, since `key2EntityMap` has no such key, but the line stays in the file:

| | requests issued | index file lines |
|---|---|---|
| run 3, `v1` removed from the graph | 0 | 1 |

So the file only ever grows, and it keeps vectors for values that no longer exist. Harmless for recall, since orphan records are not loaded, but it means the file size does not track the graph and a later entity reusing the same id would silently adopt the old vector.

*Possible directions, not a proposal yet.* The minimal change is to put a hash of the embedded text into the record and compare it during the rebuild, treating a mismatch as not indexed, which turns both symptoms into a re-embed. That leaves the index file format changed, so it needs a decision about existing files: records without a hash could be treated as unknown and re-embedded once, or as valid until the entity is next touched. Pruning orphan records wants a rewrite of the file rather than an append, which is a separate piece of work from invalidation and may be better done as compaction.

Happy to work on it once there is agreement on the record format and on how existing index files should be treated. I have the reproduction above as a runnable harness and can turn it into a test either way.

E2ern1ty:

Fixes #844.

`EmbeddingIndexStore` keyed each persisted vector by the entity key, a prefix plus the id and the label, with nothing derived from the value that was embedded. `initStore` read the presence of that key as "this entity is indexed", so an entity that kept its id and changed its value silently kept the vector of the value it no longer had. Retrieval for the old value still reached the entity while its current value was unreachable, and because the subgraph handed on is verbalized from the current graph, a query about the old value came back with a context showing the new one.

**A record now says which text produced it.** `EmbeddingResult` gains a `contentHash` field, SHA-256 over the entity's chunk list. On rebuild a record is loaded only when that fingerprint matches what the entity would be embedded from now; a mismatch counts as not indexed, so the entity is embedded again. Both paths derive the chunk list through one method, so the text hashed on the way in is the text hashed on the way out. The number of records dropped is logged, since the complaint in the issue is not only that the vector was stale but that nothing said so.

**Existing index files** are treated as the first of the two options the issue put up for decision. A record without a fingerprint cannot be shown to match the current value, and trusting it would keep exactly the staleness the check is for, so it is not loaded and its entity is embedded once. The other option, valid until the entity is next touched, never comes due, because nothing touches such a record.

The production change is +149 lines across three files, of which about 45 come from the review response below rather than from the fix itself.

### Measured against a real embedding model

The same harness was run on `master` (3f73eb55) and on this branch, both driving the store at BAAI/bge-m3, 1024 dimensions, through a local proxy that counts requests and records what is sent. The two values used are unrelated: `cosine(embed(old), embed(new)) = 0.2923`, so a hit on one is not a hit on both.

**A. The value of an entity changes, id and label unchanged.**

| | master | this PR |
| --- | --- | --- |
| requests on the second run | **0** | **1**, carrying the current value |
| cosine(vector held, embed(old value, now gone)) | **1.0000** | 0.2923 |
| cosine(vector held, embed(current value)) | **0.2923** | **1.0000** |
| query = embed(old value) reaches the entity | **yes**, and the context shows the **new** value | no |
| query = embed(current value) reaches the entity | **no** | yes, context shows the current value |

The master column reproduces the issue exactly, including the part a fact-checking layer cannot see from the outside: the entity is retrieved *because of* the old value and the reader is then shown the new one. The issue reported 1.0000 and 0.2357 for those two cosines against its own pair of values; here the same measurement gives 1.0000 and 0.2923.

**B. An entity is removed and another takes its id over**, which is the harm the issue names for records that are never pruned. Three runs: indexed, gone, id reused with a different value.

| | master | this PR |
| --- | --- | --- |
| requests on the third run | **0** | **1** |
| cosine(vector held, embed(value of the entity that is gone)) | **1.0000** | 0.2923 |
| cosine(vector held, embed(its own value)) | **0.2923** | **1.0000** |

So on master the new entity silently adopts the vector left behind. It no longer does, and that does not wait for compaction.

**C. A value the entity has held before comes back.** On this branch the third run issues **0 requests** and the file stays at 2 lines: the record already there matches again. This is what bounds the growth of an append-only file to one record set per distinct value.

**D. A record with no fingerprint**, holding a vector that happens to be correct. On master it is kept and nothing is sent. On this branch the first run sends **1 request** and the second sends **0** — the cost of the format change is paid once, per entity, and never again.

Nothing about the credentials used is in the diff; the harness read them from the environment and is not part of this PR.

### Review response, second commit

Two things were raised about `indexStoreMap`, both about the map rather than the fingerprint. Addressed in 923347fc, though not quite as suggested; the reasoning is in the review threads.

- **Keyed by `ModelUtils.getGraphEntityKey` instead of by the entity.** Lookups across instances already worked, since `GraphVertex.equals` answers on the id and the label, and what a key retained was the `Vertex` and its values rather than the graph. The reason to change it is narrower: with the entity as the key, what counts as the same indexed entity is decided by `GraphVertex.equals`, and this is a fix about an entity's identity not covering its value. Someone extending that `equals` to cover the values would silently stop `getEntityIndex` finding records and stop the scan skipping what is indexed. The file has always been keyed by the entity key; now the map is too.
- **Published once, rather than made concurrent.** `initStore` built into the field and mutated it throughout, so a reader on another thread saw an index half way through being built. A `ConcurrentHashMap` makes each operation safe but leaves that reader looking at a partial index, which on a recall path is worse than an error. So `initStore` builds into a local map and assigns it in one go, the field is `volatile` and starts empty, and a reader sees either the index that was there before or the finished one — and no longer needs a build to have happened to avoid an NPE. This does not make concurrent `initStore` calls safe against each other; they still race on the file, as before.

### What is deliberately left out

The file is still only appended to, which is where the issue puts the line: *"Pruning orphan records wants a rewrite of the file rather than an append, which is a separate piece of work from invalidation and may be better done as compaction."*

So a superseded record stays in the file and is rejected on each later read, and the records of deleted entities stay too. Measurements C and B above are why that is liveable: growth is bounded by the number of distinct values an entity has held rather than by the number of changes, and the dangerous half of the pruning problem is already closed by the fingerprint. What remains for compaction is only that the file size does not track the graph.

I had a rewrite-on-load in the first revision of this PR and took it out. It cost a public `upgradeIndexFile`, a temp-file-and-fsync write path, and the bookkeeping to keep a graph that scanned short from taking the whole file with it — about 220 lines to prune history the issue had already scoped out.

### One thing to look at: a test fixture is removed

`GraphMemoryTest` no longer loads an embedding store from `src/test/resources/index/LDBCEmbeddingIndexStore`, and that 4.6MB file is deleted. Two findings led there, both checkable before the deletion:

- It could not be carried over. Its records line up with the text the current code would embed for **14 of its 163 entities**; the other 149 disagree on chunk count, in a pattern consistent with the file having been written when the values were joined before splitting rather than split per value. So there was no honest fingerprint to stamp on it, and no way to embed it again offline.
- It could not have affected a single assertion. Every query in that test embeds to `double[0]` through `MockChatRobot`, and `EmbeddingVector.match` returns `0.0` on a length mismatch, below the `0.5` threshold. Everything asserted there comes from the keyword path, and those 22 assertions pass unchanged. Runtime for that test drops from 27.6s to 0.6s, the 27s having been model retries against a null URL.

If you would rather keep the file, I can drop only the wiring, though it would then be unreferenced. The `**/resources/index/**` rat exclusion in the root pom is now unused; I left it alone as it is a generic pattern, and can remove it on request.

### Two known limits, neither introduced here

- The fingerprint covers the text only, so changing the embedding model or its dimension leaves every record loaded, and on a dimension change recall would quietly go to zero. Same class of staleness, but which properties of a model identify it is a separate decision. Happy to open a follow-up.
- A record set short of a chunk still matches on every record it has, so an entity can be held with part of its vectors — a fingerprint covers all of an entity's chunks at once and carries no ordinal. Reachable through the blank-response skip in `indexBatch`. This is unchanged from before: a partial set was skipped as indexed then too.

The verify path also takes `verbalize(GraphEntity)` to be a function of the entity alone, noted in the javadoc where it is relied on. The one implementation in tree is.

### How was this PR tested?

Verified against a real hosted embedding model as set out above, though not in a production deployment, hence the second box.

`EmbeddingIndexInvalidationTest` covers the same ground offline, driving the store against a local `/v1/embeddings` endpoint that answers with one hot vectors, one dimension per distinct text, so a stored vector says which text produced it: cosine 1 against its own text and 0 against any other. No key and no network needed. Eight cases:

| case | what it pins |
| --- | --- |
| changed value on a vertex | embedded again, holds the current vector and not the old one, and the record just written is then accepted with the file left byte for byte |
| changed value on an edge | the same for edges |
| recall through `GraphMemoryServer` | the old value reaches nothing, the current value reaches the vertex and reads as that value |
| id taken over by another entity | the vector left behind is not adopted |
| value that comes back | served from the record already there, nothing appended |
| record without a fingerprint | not taken on trust, and the cost is paid once |
| read through an entity object from another graph | the index is found through either, and a store with no index yet holds nothing rather than failing |
| read while a build is in flight | the index that was there before, never a part of the one being built |

Each case fails against the behaviour it pins, checked by disabling the fingerprint comparison and, for the last one, by assigning the index before the build instead of after. The byte-identical assertion on the unchanged run is what would catch a fingerprint computed differently on the two paths.

`mvn -pl geaflow-ai clean test -Pjdk8`: 14 tests, 0 failures, checkstyle and RAT clean. The one case with threads in it was run repeatedly to check it settles. Nothing in `geaflow-ai/src/main` constructs this store today, so the added per-entity hashing on load has no production caller in tree.


geaflow-ai/src/main/java/org/apache/geaflow/ai/index/EmbeddingIndexStore.java
```diff
@@ -96,24 +103,45 @@ public void initStore(GraphAccessor graphAccessor, VerbalizationFunction func,
...
+                    continue;
                 }
+                String key = embedding == null ? null : embedding.input;
+                GraphEntity entity = key == null ? null : key2EntityMap.get(key);
+                if (entity != null) {
+                    if (embedding.contentHash == null) {
+                        unversionedRecords++;
+                    } else if (embedding.contentHash.equals(
+                            currentFingerprint(key, entity, key2Fingerprint))) {
+                        this.indexStoreMap.computeIfAbsent(entity, k -> new ArrayList<>()).add(embedding);
```
kitalkuyo-gita: I've observed a phenomenon: the code uses a `Map<GraphEntity, List<EmbeddingResult>>(indexStoreMap)` and populates it with entity instances in the `initStore`; however, subsequent locations can access the index using other `GraphAccessor`/`GraphEntity` instances (e.g., the `GraphEntity` instance generated when `getEntityIndex` is called externally might not be the same object reference). If the `GraphEntity` lacks appropriate `equals/hashCode` or object instance adjacency, it can lead to not finding loaded records (a logical error); furthermore, using a `GraphEntity` object as a long-term key maintains a strong reference to the graph, causing memory leaks.

Relevant code:

```
// Snippet from EmbeddingIndexStore.initStore (already exists in the PR)

GraphEntity entity = key2EntityMap.get(key);

if (entity != null) {

this.indexStoreMap.computeIfAbsent(entity, k -> new ArrayList<>()).add(embedding);

}
```

Suggested solution (prioritize stable string keys instead of objects): Change the indexStoreMap to a Map<String, List<EmbeddingResult>> with the entity key (ModelUtils.getGraphEntityKey(entity)) as the key, and replace the key with this string in all reads. This avoids memory issues caused by referencing entity objects and ensures consistency across Accessors/instances.

geaflow-ai/src/main/java/org/apache/geaflow/ai/index/EmbeddingIndexStore.java
```diff
@@ -96,24 +103,45 @@ public void initStore(GraphAccessor graphAccessor, VerbalizationFunction func,
...
                     continue;
                 }
+                EmbeddingService.EmbeddingResult embedding;
                 try {
-                    EmbeddingService.EmbeddingResult embedding =
-                            new Gson().fromJson(line, EmbeddingService.EmbeddingResult.class);
-                    String key = embedding.input;
-                    GraphEntity entity = key2EntityMap.get(key);
-                    if (entity != null) {
-                        this.indexStoreMap.computeIfAbsent(entity, k -> new ArrayList<>()).add(embedding);
```
kitalkuyo-gita, line 105: If the indexStoreMap is read and written by different threads during runtime (e.g., index initialization on the startup thread, querying on the worker thread), concurrent read/write safety must be guaranteed. Replace the map with a ConcurrentHashMap.

geaflow-ai/src/main/java/org/apache/geaflow/ai/index/EmbeddingIndexStore.java
E2ern1ty: Done in 923347fc, keyed by `ModelUtils.getGraphEntityKey` as you suggest, though I want to correct the two reasons given, because they point at the wrong risk and I would rather the record be accurate.

**Lookups across instances already worked.** `GraphVertex.equals`/`hashCode` delegate to `Vertex`, which is `Objects.equals(id) && Objects.equals(label)`, and `GraphEdge` to `Edge`, which is `srcId`, `dstId` and `label`. So an instance handed out by a different `GraphAccessor` for the same entity is equal and hashes alike, and `getEntityIndex` found the records. I could not construct a case on master where it did not.

**What a key retained was not the graph.** `GraphVertex` holds a `Vertex`, and `Vertex` holds `id`, `label` and `values` — no reference to `MemoryGraph`, `GraphAccessor` or anything above it. So the retention was of the `Vertex` objects and their value strings, which is real but bounded by the entity count and small beside the vectors in the same map, at 1024 doubles each. Not a leak in the sense of holding a graph alive.

**The reason I think the change is right is narrower, and closer to what this PR is about.** With the entity as the key, what counts as the same indexed entity is decided by `GraphVertex.equals`, which answers on the id and the label. This PR exists because an entity's identity did not cover its value. Extending `Vertex.equals` to cover the values is a reasonable thing for someone to want next, and the moment they did, `getEntityIndex` would stop finding records and the scan in `initStore` would stop skipping what is already indexed — silently, both of them. Meanwhile the file has always been keyed by the entity key. Two notions of identity for one thing, one of them defined in a class this PR invites people to change. Keying the map by the same string as the file collapses them into one and takes the index out of reach of that change.

The call sites moved with it: the scan loop and the in-batch dedup set now hold keys, `indexBatch` writes under the key it already computes for `res.input`, and `getEntityIndex` looks up `getGraphEntityKey(entity)`. That last one builds a small string per lookup where before it hashed two fields, which on the query path is one allocation per vertex per query, against a 1024-dimension cosine for each of them.

There is a test for the scenario you describe: it indexes through one graph and reads through an entity object from another, asserting first that the two objects really are different. It holds on master as well — it is there to keep it holding.

geaflow-ai/src/main/java/org/apache/geaflow/ai/index/EmbeddingIndexStore.java
E2ern1ty: Addressed in 923347fc, but not with a `ConcurrentHashMap`, because I do not think that fixes the case you describe.

Take it as stated: initialization on the startup thread, a query on a worker thread. `initStore` assigned the map to the field up front and then mutated it for the whole of its run — the file read, then batch after batch of embedding. A worker reading in that window sees whatever has been filled in so far. With a `ConcurrentHashMap` it still does: every individual `get` is safe, no exception is thrown, and the answer is a partial index. For a recall path that is worse than an error, because it comes back as a plausible result rather than a failure. The same goes for a second `initStore` on a live store, which under either map empties the index for readers before filling it again.

So the map is built aside and assigned to the field in one go, at the end of `initStore`. The field is `volatile`, and starts as `Collections.emptyMap()`. A reader now sees either the index that was there before or the finished one, and a reader that arrives before any build has happened gets an empty list rather than the `NullPointerException` it would have got on master. `getEntityIndex` reads the field once per call, so it works against one map throughout.

Two notes on the trade-off, since it is not free:

- During a rebuild both maps are held, so peak memory is roughly doubled for that window. I think consistency is worth it here, and there is no caller in tree that rebuilds a live store — nothing in `geaflow-ai/src/main` constructs this store at all today.
- This does not make `initStore` safe to call concurrently with itself. Two builds at once still race on the index file, which is appended to. That was true before and I have not tried to fix it here; if you would like it enforced, the smallest thing would be to reject a concurrent call outright rather than to make it work.

There is a test: it holds the embeddings endpoint open with a latch, runs a second `initStore` on the same store on another thread, and asserts that a reader in that window gets the vector from before the build. It fails if the field is assigned before the build instead of after — with the assignment moved back, the reader gets zero vectors, which is the partial index a `ConcurrentHashMap` would also have shown.

